PR #9, opened 2018-08-29
Remove default fallbacks and add Tables.istable interface check function

### LINKED ISSUE #5: Default implementations of methods in Tables.jl
I started to test the package and I have one thing that was not clear to me.
Why do you provide default implementations for the methods in this package, e.g. `rows` and `columns`, `schema`. Users should implement a specific method for them anyhow if they are to be used so maybe it would be better if by default they thrown "not imlemented" error?

(I am starting to digest the package so I might be missing something)

quinnj:
cc @bkamins, should resolve #5

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -85,11 +85,16 @@ abstract type Table end
 abstract type AccessStyle end
 struct RowAccess <: AccessStyle end
 struct ColumnAccess <: AccessStyle end
-AccessStyle(x) = RowAccess()
 
 "Tables.schema(s) => NamedTuple{names, types}"
 function schema end
-schema(x) = eltype(x)
+
+function istable(x::T) where {T}
+    hasmethod(AccessStyle, Tuple{T}) &&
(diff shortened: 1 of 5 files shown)


src/Tables.jl
```diff
@@ -85,11 +85,16 @@ abstract type Table end
...
-AccessStyle(x) = RowAccess()
 
 "Tables.schema(s) => NamedTuple{names, types}"
 function schema end
-schema(x) = eltype(x)
+
+function istable(x::T) where {T}
+    hasmethod(AccessStyle, Tuple{T}) &&
+    hasmethod(schema, Tuple{T}) &&
+    AccessStyle(T) === RowAccess() ? hasmethod(rows, Tuple{T}) :
```
bkamins, line 95: Very nice! I have only the question if we need the tests `hasmethod(rows, Tuple{T})` in this line and `hasmethod(columns, Tuple{T})` in line below. I guess they will always evaluate to `true` as `rows` and `columns` functions accept any `x` in definitions below.

Issue #7, opened 2018-08-27
TableTraits.jl/Query.jl integration issues


davidanthoff:
I just had a quick look through the integration, and here are some problems I spotted:
1. [This](https://github.com/JuliaData/Tables.jl/blob/4eca9f0dec84d428bc3b5029db9f6bba6badd4ee/src/enumerable.jl#L29) line will break if there is a case of ``EltypeUnknown``, I believe? This is one of the new changes for the 0.7/1.0 version of TableTraits.jl: an iterable table can return ``EltypeUnknown``, and a sink needs to handle that.
2. In general I think this package should not assume anything about ``Enumerable`` or depend on it, that is really an implementation detail. The proper way to consume a query result is to just follow the iterable tables interface for consuming it. The current version might work right now for some queries, but it won't work for any other Query.jl backend, it doesn't work for all the other TableTraits.jl sources there are (all the file IO etc.), and in general one should treat all the ``Enumerable`` types as non-public and not rely on their existence (i.e. all details about that part of Query.jl might break at any moment).
3. How would this turn a Tables.jl source into an iterable table? Would folks have to manually define ``getiterator`` etc.? But then they have to take a dependency on the Queryverse stuff, which I think you wanted to avoid? I didn't look very long, so maybe I missed something there?

I haven't really reviewed this carefully, so there might be more stuff coming :) I also have very little time the next 1-2 weeks, semester started and that is always a tricky phase.

quinnj:
1. This is a fair point and one that I've definitely pondered over; the conclusion I came to for the moment is that I'm not actually aware of any cases where sources will return `EltypeUnknown`, particularly in the table world. I certainly understand the use-case in the more general Query world where you might be dealing w/ all sorts of kinds of iterators, but at least in the table world, I wonder if there actually would be a case where a table source didn't know its own schema? I remember seeing this change in TableTraits, but how is it actually planned to be handled? Does each sink have to account for that case specifically? Would you provide some kind of helper function to "collect" that case? It's certainly something I'd like to keep discussing/pondering, but also not an urgent issue as far as I'm aware.
2. Hmmm, I didn't realize `Enumerable` was such an implementation detail. The main idea here was just to find some kind of "post-hook" for tables coming out of Query/QueryOperators where we could do the automatic DataValue-unwrapping. For now, this works quite well for those cases specifically, but we'll have to think through a more robust solution here.
[...]

PR #10, opened 2018-08-31
Support unknown schema from sources

quinnj:
@davidanthoff @piever

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,11 @@ Where `Rows` and `Columns` are the duals of each other:
 * `Columns` is a property-accessible object of iterators (i.e. each column is an iterator)
 
 In addition to these `Rows` and `Columns` objects, it's useful to be able to query properties of these objects:
-* `Tables.schema(rows_or_columns)`: returns a `Tables.Schema` object
+* `Tables.schema(x::Union{Rows, Columns}) => Union{Tables.Schema, Nothing}`: returns a `Tables.Schema` object, or `nothing` if the table's schema is unknown
+* For the `Tables.Schema` object:
   * column names can be accessed as a tuple of Symbols like `sch.names`
-  * column types can be accessed as a tuple of types like `sch.types`, though note that `sch.types` may return `nothing`, indicating column types are not computable or otherwise unknown
+  * column types can be accessed as a tuple of types like `sch.types`
+  * See `?Table.Schema` for more details on this type
 
 A big part of the power in these simple interface functions is that each (`Tables.rows` & `Tables.columns`) is defined for any table type, even if the table type only explicitly implements one interface function or the other.
 This is accomplished by providing performant, generic fallback definitions in Tables.jl itself (though obviously nothing prevents a table type from implementing each interface function directly).
@@ -24,17 +26,18 @@ With these simple definitions, powerful workflows are enabled:
 
 So how does one go about satisfying the Tables.jl interface functions? It mainly depends on what you've alrady defined and the natural access patterns of your table:
 
+First:
+* `Tables.istable(::Type{<:MyTable}) = true`: this provides an explicit affirmation that your type implements the Tables interface
+
 To support `Rows`:
-* Define `Tables.rowaccess(::Type{MyTable}) = true`: this signals to other types that `MyTable` supports valid `Row`-iteration
+* Define `Tables.rowaccess(::Type{<:MyTable}) = true`: this signals to other types that `MyTable` supports valid `Row`-iteration
 * Define `Tables.rows(x::MyTable)`: return a `Row`-iterator object (perhaps the table itself if already defined)
+* Define `Tables.schema(Tables.rows(x::MyTable))` to either return a `Tables.Schema` object, or `nothing` if the schema is unknown or non-inferrable for some reason
 
 To support `Columns`:
-* Define `Tables.columnaccess(::Type{MyTable}) = true`: this signals to other types that `MyTable` supports returning a valid `Columns` object
+* Define `Tables.columnaccess(::Type{<:MyTable}) = true`: this signals to other types that `MyTable` supports returning a valid `Columns` object
 * Define `Tables.columns(x::MyTable)`: return a `Columns`, property-accessible object (perhaps the table itself if it naturally supports property-access to columns)
-
-In addition, it can be helpful to define the following:
-* `Tables.istable(::Type{MyTable}) = true`: there are runtime checks to see if an object implements the interface, but this can provide an explicit affirmation
-* `Tables.schema(rows_or_columns)`: allow users to get the column names & types of `MyTable`'s `Rows` or `Columns` objects as a tuple of Symbols and tuple of types; by default, it will call `propertynames` on the first row or `Columns` object directly for names; column types will be `nothing` by default
+* Define `Tables.schema(Tables.columns(x::MyTable))` to either return a `Tables.Schema` object, or `nothing` if the schema is unknown or non-inferrable for some reason
 
 The final question is how `MyTable` can be a "sink" for any other table type. The answer is quite simple: use the interface functions!
 
@@ -54,7 +57,7 @@ function MyTable(x)
     for row in rows
         # a convenience function provided in Tables.jl for "unrolling" access to each column/property of a `Row`
         # it works by applying a provided function to each value; see `?Tables.eachcolumn` for more details
-        Tables.eachcolumn(sch, row, mytbl) do val, col, name, mytbl
+        Tables.eachcolumn(sch, row) do val, col, name
             push!(mytbl[col], val)
         end
     end
@@ -68,7 +71,7 @@ function MyTable(x)
     Tables.istable(x) || throw(ArgumentError("MyTable requires a table input"))
     cols = Tables.columns(x)
     # here we use Tables.eachcolumn to iterate over each column in a `Columns` object
-    return MyTable(collect(Tables.names(cols)), [collect(col) for col in Tables.eachcolumn(cols)])
+    return MyTable(collect(propertynames(cols)), [collect(col) for col in Tables.eachcolumn(cols)])
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -1,68 +1,75 @@
 ### Tables.jl
 
-The Tables.jl package provides four useful interface functions for working with tabular data in a variety of formats.
+The Tables.jl package provides simple, yet powerful interface functions for working with all kinds tabular data through predictable access patterns.
 
 ```julia
-    Tables.schema(table) => NamedTuple{names, types}
-    Tables.AccessStyle(table_type) => Tables.RowAccess() || Tables.ColumnAccess()
-    Tables.rows(table) => iterator with values accessible via getproperty(row, columnname)
-    Tables.columns(table) => Collection of iterators, with each column accessible via getproperty(x, columnname)
+    Tables.rows(table) => Rows
+    Tables.columns(table) => Columns
 ```
+Where `Rows` and `Columns` are the duals of each other:
+* `Rows` is an iterator of property-accessible objects (any type that supports `propertynames(row)` and `getproperty(row, nm::Symbol`)
+* `Columns` is a property-accessible object of iterators (i.e. each column is an iterator)
 
-Essentially, for any table type that implements the interface requirements, one can get access to:
+In addition to these `Rows` and `Columns` objects, it's useful to be able to query properties of these objects:
+* `Tables.schema(rows_or_columns)`: returns a `Tables.Schema` object
+  * column names can be accessed as a tuple of Symbols like `sch.names`
+  * column types can be accessed as a tuple of types like `sch.types`, though note that `sch.types` may return `nothing`, indicating column types are not computable or otherwise unknown
 
-1. the schema of a table (it's column names & element types), returned as a NamedTuple _type_ (`NamedTuple{names, types}`), with parameters of `names` which are the column names as a Tuple of Symbols, and `types` which is a Tuple type of types (`Tuple{...}`)
-2. rows of the table via a Row-iterator
-3. the columns of the table via a collection of iterators, where the individual column iterators are accessible via getproperty(table, columnname)
+A big part of the power in these simple interface functions is that each (`Tables.rows` & `Tables.columns`) is defined for any table type, even if the table type only explicitly implements one interface function or the other.
+This is accomplished by providing performant, generic fallback definitions in Tables.jl itself (though obviously nothing prevents a table type from implementing each interface function directly).
 
-Let's clarify the definitions of some of the terms that were just used:
+With these simple definitions, powerful workflows are enabled:
+* A package providing data cleansing, manipulation, visualization, or analysis can automatically handle any number of decoupled input table types
+* A tabular file format can have automatic integration with in-memory structures and translation to other file formats
 
-* `Tables.rows(src)` returns a `Row` iterator, where `Row` is any object that supports value access via `getproperty(row, nm::Symbol)`. For example, if I have a NamedTuple like `row = (a=1, b=2, c=3)`, its values can be accessed like `row.a`, which is desugared to a call to `getproperty(row, :a)`. Thus a NamedTuple implicitly satisfies the `Row` interface. A `Row` should support `getproperty(row, name)` for any column name returned in a table's `Tables.schema(src)`. This allows an end-user access pattern like:
+So how does one go about satisfying the Tables.jl interface functions? It mainly depends on what you've alrady defined and the natural access patterns of your table:
 
+To support `Rows`:
+* Define `Tables.rowaccess(::Type{MyTable}) = true`: this signals to other types that `MyTable` supports valid `Row`-iteration
+* Define `Tables.rows(x::MyTable)`: return a `Row`-iterator object (perhaps the table itself if already defined)
+
+To support `Columns`:
+* Define `Tables.columnaccess(::Type{MyTable}) = true`: this signals to other types that `MyTable` supports returning a valid `Columns` object
+* Define `Tables.columns(x::MyTable)`: return a `Columns`, property-accessible object (perhaps the table itself if it naturally supports property-access to columns)
+
+In addition, it can be helpful to define the following:
+* `Tables.istable(::Type{MyTable}) = true`: there are runtime checks to see if an object implements the interface, but this can provide an explicit affirmation
+* `Tables.schema(rows_or_columns)`: allow users to get the column names & types of `MyTable`'s `Rows` or `Columns` objects as a tuple of Symbols and tuple of types; by default, it will call `propertynames` on the first row or `Columns` object directly for names; column types will be `nothing` by default
+
+The final question is how `MyTable` can be a "sink" for any other table type. The answer is quite simple: use the interface functions!
+
+* Define a function or constructor that takes, at a minimum, a single, untyped argument and then calls `Tables.rows` or `Tables.columns` on that argument to construct an instance of `MyTable`
+
+For example, if `MyTable` is a row-oriented format, I might define my "sink" function like:
 ```julia
-sch = Tables.schema(src)
-names = Tables.names(sch)
-for row in Tables.rows(src)
-    for nm in names
-        foo(getproperty(row, nm))
-    end
-end
-```
-* `Tables.columns(src)` returns an object that satisifes the `Columns` interface, which, similar to a `Row`, allows access to values via `getproperty(columns, nm::Symbol)`, but, unlike a `Row` returning a scalar value, will return a vector or iterator representing all the values in the column. 
-
-
-So how does one go about satisfying these interface functions as a table type developer? It mainly depends on the `Tables.AccessStyle(T)` of your table:
-
-* `Tables.schema(x)`: given an _instance_ of a table type, generate a `NamedTuple` type, with a tuple of symbols for column names (e.g. `(:a, :b, :c)`), and a tuple type of types as the 2nd parameter (e.g. `Tuple{Int, Float64, String}`); like `NamedTuple{(:a, :b, :c), Tuple{Int, Float64, String}}`
-
-* `Tables.RowAccess()`:
-  * overload `Tables.rows` for your table type (e.g. `Tables.rows(t::MyTableType)`), and return an iterator of `Row`s. Where a `Row` type is any object with keys accessible via `getproperty(obj, key)` (like a NamedTuple)
-
-* `Tables.ColumnAccess()`:
-  * overload `Tables.columns` for your table type (e.g. `Tables.columns(t::MyTableType)`), returning a collection of iterators, with individual column iterators accessible via column names by `getproperty(x, columnname)` (a NamedTuple of Vectors, for example, would satisfy the interface)
-
-The final question is how `MyTableType` can be a "sink" for any other table type:
-
-* Define a function or constructor that takes, at a minimum, a single, untyped argument and then uses the Tables.jl interface functions (`Tables.schema`, `Tables.rows` and `Tables.columns`) on that argument to construct an instance of `MyTableType`
-
-For example, if `MyTableType` is a row-oriented format, I might define my "sink" function like:
-```julia
-function MyTableType(x)
-    mytbl = MyTableType(Tables.schema(x)) # custom constructor that creates an "empty" MyTableType w/ the right schema
-    for row in Tables.rows(x)
-        append!(mytbl, row)
+function MyTable(x)
+    Tables.istable(x) || throw(ArgumentError("MyTable requires a table input"))
+    rows = Tables.rows(x)
+    sch = Tables.schema(rows)
+    names = sch.names
+    types = sch.types
+    # custom constructor that creates an "empty" MyTable according to given column names & types
+    # note that the "unknown" schema case should be considered, i.e. when `sch.types => nothing`
+    mytbl = MyTable(names, types)
+    for row in rows
+        # a convenience function provided in Tables.jl for "unrolling" access to each column/property of a `Row`
+        # it works by applying a provided function to each value; see `?Tables.eachcolumn` for more details
+        Tables.eachcolumn(sch, row, mytbl) do val, col, name, mytbl
+            push!(mytbl[col], val)
+        end
     end
     return mytbl
 end
 ```
 
-Alternatively, if `MyTableType` is column-oriented, perhaps my definition would be more like:
-
+Alternatively, if `MyTable` is column-oriented, perhaps my definition would be more like:
 ```julia
-function MyTableType(x)
+function MyTable(x)
+    Tables.istable(x) || throw(ArgumentError("MyTable requires a table input"))
     cols = Tables.columns(x)
-    return MyTableType(collect(map(String, keys(cols))), [col for col in cols])
+    # here we use Tables.eachcolumn to iterate over each column in a `Columns` object
+    return MyTable(collect(Tables.names(cols)), [collect(col) for col in Tables.eachcolumn(cols)])
 end
 ```
 
-Obviously every table type is different, but via a combination of `Tables.schema`, `Tables.rows`, and `Tables.columns`, each table type should be able to construct an instance of itself.
+Obviously every table type is different, but via a combination of `Tables.rows` and `Tables.columns` each table type should be able to construct an instance of itself.
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -31,86 +31,156 @@ function __init__()
     end
 end
 
-include("utils.jl")
-
 "Abstract row type with a simple required interface: row values are accessible via `getproperty(row, field)`; for example, a NamedTuple like `nt = (a=1, b=2, c=3)` can access its value for `a` like `nt.a` which turns into a call to the function `getproperty(nt, :a)`"
 abstract type Row end
 
 """
-The Tables.jl package provides four useful interface functions for working with tabular data in a variety of formats.
+The Tables.jl package provides simple, yet powerful interface functions for working with all kinds tabular data through predictable access patterns.
 
 ```julia
-    Tables.schema(table) => NamedTuple{names, types}
-    Tables.AccessStyle(table_type) => Tables.RowAccess() || Tables.ColumnAccess()
-    Tables.rows(table) => iterator with values accessible via getproperty(row, columnname)
-    Tables.columns(table) => Collection of iterators, with each column accessible via getproperty(x, columnname)
+    Tables.rows(table) => Rows
+    Tables.columns(table) => Columns
 ```
+Where `Rows` and `Columns` are the duals of each other:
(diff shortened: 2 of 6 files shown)


quinnj:
This is getting in pretty good shape; there are still some stray allocations coming from somewhere, perhaps the getproperty isn't getting inlined or something. It's about a factor of ~3x of the known schema case.

quinnj:
@davidanthoff @piever @bkamins @andyferris @nalimilan ,

This PR now represents a slightly larger refactoring than anticipated of the Tables.jl interfaces. In particular, I'm proposing the following changes:
* Insead of AccessStyle traits, use `Tables.rowaccess(::Type{MyType}) = true` overloads to affirm a table supports a certain interface access pattern; I think this this overall simplifies things because tables can just signal/opt-in to supporting a certain interface that generic fallbacks or other users can check later on. In the same vein as `Tables.istable(T)`, these just live at the individual interface function level
* Introduce a new `Tables.Schema{names, T}` type; this actually isn't drastically different than the `Tables.schema(x)` pattern before, but we need this custom type to better handle the case of unknown column types, since a `NamedTuple{names, types}` requires names and types to be the same length. Now a table w/ an unknown schema will have a schema like `Tables.Schema{names, nothing}`, with `nothing` indicating that column types are not known. I opted to use `nothing` instead of `missing` since we don't want any propagation of the missingness here and users/developers will need to handle that case specially with explicit checks (though convenience functions are provided like `Tables.eachcolumn` that will adjust for users).
* The other change also tries to further reduce the "schema" role in the Tables.jl story by only having `Tables.schema(x)` be called on the result of `Tables.rows(x)` or `Tables.columns(x)`; this makes the call more likely to succeed if a table type returns a separate object w/ more name/type information, and also allows for easier use where users/developers don't actually need to ever use the schema.

quinnj:
(I've updated the DataFrames PR and put up a CSV [PR](https://github.com/JuliaData/CSV.jl/pull/262) that both conform to these changes for reference)

src/Tables.jl
```diff
@@ -148,34 +216,89 @@ end
...
+        copyto!(new, dest)
+        push!(new, el)
+        return new
+    end
+end
+
+# when Tables.types(x) === nothing
+function buildcolumns(sch::Schema{names, nothing}, rowitr::T) where {names, T}
+    state = iterate(rowitr)
+    state === nothing && return NamedTuple{names}(Tuple(Missing[] for _ in names))
```
piever: IndexedTables uses inference in this case to guess the type, though returning `Missing` type like you do here is probably more principled. We should be consistent, but maybe it's the IndexedTables side that needs to change?

src/Tables.jl
```diff
@@ -148,34 +216,89 @@ end
...
+    row::eltype(rowitr), st = state
+    cols = length(names)
+    L = Base.IteratorSize(T)
+    len = haslength(L) ? length(rowitr) : 0
+    columns = Vector{AbstractVector}(undef, cols)
```
piever: One of the things that got me stuck when trying to implement this in this framework is that it's not completely clear what structure should contain the columns. `NamedTuple` is out of the question because they can change type, but doesn't a `Vector{AbstractVector}` cause performance issues?

piever: I was even thinking, maybe the only performant way to do this is to use generated functions that create code where we use a different variable per each column, but maybe that's overkill.

src/utils.jl
```diff
@@ -1,49 +1,49 @@
...
 
-    Takes a function `f`, a NamedTuple type `schema`, a `row` type (that satisfies the Row interface), and any other `args...`;
-    it generates calls to get the value for each column in the row and then calls `f(val, T, col, name)`, where `f` is the
-    user-provided function, `val` is a row's column value, `T` is the column's element type, `col` is the column index as
-    an `Int`, and `name` is the row's column name.
-
-    This is useful for sinks iterating rows who wish to provide a type-stable mechanism for their "inner loops". Typically,
-    such inner loops suffer from dynamic dispath due to the varying types of columns.
+    Takes a function `f`, column names `names`, column types `types`, a `row` type (that satisfies the Row interface), and any other `args...`;
+    it generates calls to get the value for each column in the row (`getproperty(row, nm)`) and then calls `f(val, col, name)`, where `f` is the
```
piever: `f(val, col, name, args...)`, right?

piever:
Looks really nice. I've left a couple of comments, but I need to think a bit more carefully how this relates to what we have in StructArrays and IndexedTables. In particular I'm curious about nesting. Say I have an iterator of `Pair{NamedTuple, NamedTuple}` like `(a = 1, b = 2) => (c=3, d = 4)` (this is actually quite useful in IndexedTables, the first `NamedTuple` is the primary keys and the other the non-primary keys. Would there be a way to use this code / framework to collect into a `Pair(NamedTuple{Vector...}, NamedTuple{Vector...})` (syntax is not quite right but I hope it's clear)?

README.md
```diff
@@ -1,68 +1,75 @@
...
+This is accomplished by providing performant, generic fallback definitions in Tables.jl itself (though obviously nothing prevents a table type from implementing each interface function directly).
 
-Let's clarify the definitions of some of the terms that were just used:
+With these simple definitions, powerful workflows are enabled:
+* A package providing data cleansing, manipulation, visualization, or analysis can automatically handle any number of decoupled input table types
+* A tabular file format can have automatic integration with in-memory structures and translation to other file formats
 
-* `Tables.rows(src)` returns a `Row` iterator, where `Row` is any object that supports value access via `getproperty(row, nm::Symbol)`. For example, if I have a NamedTuple like `row = (a=1, b=2, c=3)`, its values can be accessed like `row.a`, which is desugared to a call to `getproperty(row, :a)`. Thus a NamedTuple implicitly satisfies the `Row` interface. A `Row` should support `getproperty(row, name)` for any column name returned in a table's `Tables.schema(src)`. This allows an end-user access pattern like:
```
nalimilan, line 20: The concrete example in this paragraph was useful IMHO.

README.md
```diff
@@ -1,68 +1,75 @@
...
+* Define `Tables.rowaccess(::Type{MyTable}) = true`: this signals to other types that `MyTable` supports valid `Row`-iteration
+* Define `Tables.rows(x::MyTable)`: return a `Row`-iterator object (perhaps the table itself if already defined)
 
+To support `Columns`:
+* Define `Tables.columnaccess(::Type{MyTable}) = true`: this signals to other types that `MyTable` supports returning a valid `Columns` object
+* Define `Tables.columns(x::MyTable)`: return a `Columns`, property-accessible object (perhaps the table itself if it naturally supports property-access to columns)
 
-So how does one go about satisfying these interface functions as a table type developer? It mainly depends on the `Tables.AccessStyle(T)` of your table:
+In addition, it can be helpful to define the following:
+* `Tables.istable(::Type{MyTable}) = true`: there are runtime checks to see if an object implements the interface, but this can provide an explicit affirmation
```
nalimilan: I'd go as far as making this mandatory. Currently, checking whether a type implements a method is slow AFAIK, and it's cleaner to clearly require types to implement this trait. Else nobody can really rely on it, which makes it not that useful.

src/Tables.jl
```diff
@@ -31,110 +31,178 @@ function __init__()
...
 abstract type Table end
 
-abstract type AccessStyle end
-struct RowAccess <: AccessStyle end
-struct ColumnAccess <: AccessStyle end
-
-"Tables.schema(s) => NamedTuple{names, types}"
-function schema end
+# default definitions
+rowaccess(x) = false
```
nalimilan: Maybe limit this to `::Type`?

src/Tables.jl
```diff
@@ -31,110 +31,178 @@ function __init__()
...
+    if field === :names
+        return names
+    elseif field === :types
+        return types === nothing ? nothing : Tuple(fieldtype(types, i) for i = 1:fieldcount(types))
+    else
+        throw(ArgumentError("unsupported property for Tables.Schema"))
+    end
 end
 
+function schema(x::T) where {T}
```
nalimilan: Are fallbacks like this really useful? I can see a few reasons not to provide such a function:
- some types may not allow calling `first(x)` more than once, making this fallback incorrect
- it's quite trivial to implement for a specific type
- better get a `MethodError` than an empty schema for types that don't implement the interface

EDIT: see also https://github.com/JuliaData/DataFrames.jl/pull/1495#discussion_r214529536

src/Tables.jl
```diff
@@ -148,34 +216,89 @@ end
...
-        vals = Tuple(:(allocatecolumn($(fieldtype(NT, i)), len)) for i = 1:fieldcount(NT))
+        vals = Tuple(:(allocatecolumn($(fieldtype(types, i)), len)) for i = 1:fieldcount(types))
         return :(NamedTuple{names}(($(vals...),)))
     else
-        return NamedTuple{names}(Tuple(allocatecolumn(fieldtype(NT, i), len) for i = 1:fieldcount(NT)))
+        return NamedTuple{names}(Tuple(allocatecolumn(fieldtype(types, i), len) for i = 1:fieldcount(types)))
     end
 end
 
+haslength(T) = T === Base.HasLength() || T === Base.HasShape{1}()
```
nalimilan: It's weird to use `T` here since it's an instance rather than a type.

src/Tables.jl
nalimilan: Why `Missing[]` rather than `Union{}`? Anyway it would make sense to use inference just like `map`.

src/Tables.jl
```diff
@@ -148,34 +216,89 @@ end
...
+    state = iterate(rowitr)
+    state === nothing && return NamedTuple{names}(Tuple(Missing[] for _ in names))
+    row::eltype(rowitr), st = state
+    cols = length(names)
+    L = Base.IteratorSize(T)
+    len = haslength(L) ? length(rowitr) : 0
+    columns = Vector{AbstractVector}(undef, cols)
+    eachcolumn(add_or_widen!, sch, row, L, columns, 1, true, len)
+    rownbr = 2
+    while true
```
nalimilan: You should be able to use a `for` loop here, maybe with `outer rownbr`.

src/Tables.jl
nalimilan: Good point. Indeed, since the loop over columns is applied for each row, it's essential to avoid any type uncertainty.

Maybe we can avoid a generated function though, and instead use the same strategy as `map`: store columns as a named tuple, and have the function call itself with the updated named tuple each time the types change.

src/namedtuples.jl
```diff
@@ -28,47 +35,62 @@ function Base.iterate(rows::NamedTupleIterator{NT}, st=()) where {NT <: NamedTup
...
+columns(x::ColumnTable) = x
+schema(x::T) where {T <: ColumnTable} = Schema(names(T), types(T))
 
 _eltype(::Type{A}) where {A <: AbstractVector{T}} where {T} = T
-schema(ct::T) where {T <: ColumnTable} = schema(T)
-Base.@pure function schema(::Type{NT}) where {NT <: NamedTuple{names, T}} where {names, T <: NTuple{N, AbstractVector{S} where S}} where {N}
-    TT = Tuple{Any[ _eltype(fieldtype(NT, i)) for i = 1:fieldcount(NT) ]...}
-    return NamedTuple{names, TT}
+Base.@pure function types(::Type{NT}) where {NT <: NamedTuple{names, T}} where {names, T <: NTuple{N, AbstractVector{S} where S}} where {N}
+    return Tuple{Any[ _eltype(fieldtype(NT, i)) for i = 1:fieldcount(NT) ]...}
```
nalimilan: Use a generator instead of an array comprehension to avoid an allocation? Could also use `ntuple`, which often gives cleaner code IIRC.

src/utils.jl
```diff
@@ -1,49 +1,49 @@
...
+    push!(rle, (curT, len))
     return rle
 end
 
 # generic fallback from getproperty w/ type information to basic symbol lookup
 Base.getproperty(x, ::Type{T}, i::Int, nm::Symbol) where {T} = getproperty(x, nm)
 
 """
-    Tables.unroll(f, schema, row, args...)
+    Tables.eachcolumn(f, names, types, row, args...)
```
nalimilan: AFAICT the actual signature is `eachcolumn(f, schema, row, args...)`.

README.md
```diff
@@ -1,68 +1,75 @@
...
+    sch = Tables.schema(rows)
+    names = sch.names
+    types = sch.types
+    # custom constructor that creates an "empty" MyTable according to given column names & types
+    # note that the "unknown" schema case should be considered, i.e. when `sch.types => nothing`
+    mytbl = MyTable(names, types)
+    for row in rows
+        # a convenience function provided in Tables.jl for "unrolling" access to each column/property of a `Row`
+        # it works by applying a provided function to each value; see `?Tables.eachcolumn` for more details
+        Tables.eachcolumn(sch, row, mytbl) do val, col, name, mytbl
```
nalimilan: It's actually not needed to pass `mytbl`, and I find it confusing.

src/Tables.jl
```diff
@@ -31,110 +31,178 @@ function __init__()
...
+
+The actual type definition is
+```julia
+struct Schema{names, types} end
+```
+Where `names` is a tuple of Symbols, and `types` is a tuple _type_ of types (like `Tuple{Int64, Float64, String}`).
+Encoding the names & types as type parameters allows convenient use of the type in generated functions
+and other optimization use-cases.
+"""
+struct Schema{names, types} end
```
bkamins: A small comment: current implementation of `Schema` differs from `NamedTuple` also because it allows duplicate column names (as opposed to `NamedTuple`). Not sure if it is a significant thing.

src/Tables.jl
piever: That's the strategy in IndexedTables. However here it's a bit tricky to implement for technical reasons as in IndexedTables you are iterating over `NamedTuples` so you can tell "statically" whether you need to expand or not, just by looking at the type of the row. One strategy I was proposing [here](https://github.com/JuliaData/Tables.jl/issues/7#issuecomment-417658109) is to store columns as a `NamedTuple`, and change `add!` so that if the types are incompatible, it would return `nothing` and not do anything. Then in turn `eachcolumns` would know that if a function call returns `nothing`, it has to exit and signal that it failed, so the main loop would just call `eachcolumns` and only call itself with the updated named tuple when `eachcolumn` signals that it failed.

davidanthoff:
I didn't look through the whole PR (I'll try to find some time, but who knows right now). But this caught me eye:

> Now a table w/ an unknown schema will have a schema like Tables.Schema{names, nothing}, with nothing indicating that column types are not known.

Does that also support the case that the column names are not known? That is what we need to handle for the Query.jl situation.

src/Tables.jl
quinnj: There is no "inference" here, we're not applying a function, we're just getting an empty iterator. I considered `Union{}`, but it felt to Base-collect-y to me: here we're talking tables in the data world, where `Missing` seems a more natural default, like an all `NULL` column in a database.

src/utils.jl
quinnj: yes

src/Tables.jl
nalimilan: What I'm worried about with `Missing` is that if you later combine the resulting table with another (non-empty) one, you'll get a `Union{T,Missing}` column, which wouldn't really be appropriate.

quinnj:
After some interface soul-searching over the last few days, I think I boiled it down to this core idea: Tables.jl is really the union of two interfaces in Base: Iterable and what I'm calling PropertyAccessible (i.e. implements `propertynames` and `getproperty`). As mentioned in the docs, `Tables.rows` and `Tables.columns` are really the duals of each other with regards to these two interfaces: `Tables.columns(x)` returns a PropertyAccessible object of iterators, and `Tables.rows(x)` returns an Iterator of PropertyAccessible objects. So being able to get the "column names" (i.e. `propertynames`) of the results of `Tables.columns`  and individual `Row`-iterations from `Tables.rows` is fundamental.

Now, I realize that this isn't quite what's implemented here as you pointed out. We shouldn't ask developers to implement `Tables.schema` on the direct result of `Tables.rows(x)`, but rather on the iterated _values_ of `Tables.rows` (i.e. each "row" instance should allow calling `Tables.schema` on it). I can make that update and I don't think it changes things too much (looking over CSV.write, for example, we would just do an initial iteration, then call `Tables.schema`, then handle things appropriately.

But, I think we still need to _allow_ a user to call `Tables.schema` (or perhaps just `Tables.names`) on the table type, because a table type might have a known schema, but not have any values. I.e. `DataFrame(a=Int[], b=Float64[])`. In this case, the DataFrame knows its column names and types, yet `Tables.rows(df)` would product an empty iterator. So I think we need to somehow allow the user to call `Tables.schema` on whatever is returned by `Tables.rows` or `Tables.columns`, but somehow also signal that they should only call that in the "empty" case. This is somewhat our dual of the "what's the default type of an empty Generator for collect" problem. If a table is empty, we should have a way to say, "hey, do you happen to know your column names & types anyway?" and if so, the sink can use those, otherwise, return the equivalent of a `DataFrame()`.

src/Tables.jl
quinnj: Hmmmm, that's a good point; that certainly makes a case for `Union{}` instead.

quinnj:
Thinking about it even further, I think we _do_ still allow users to call `Tables.schema` on the direct `Tables.rows` or `Tables.columns` objects, but _also_ allow them to return `nothing` directly instead of a `Tables.Schema` object. The difference here is that Some tables know statically their column names/types and sinks should be able to use that information in the nice case. But if a table _doesn't_ know them, the sink can still rely on calling `Tables.schema` on the individual rows, but also realizing that the `Tables.Schema` may change from iteration to iteration (i.e. `(a=1, b=1.0)`, then `(a=2, b=missing)`).

I'll try to iron this out some more tomorrow morning by tweaking the implementations.

src/Tables.jl
andyferris: You possibly want an iterator (edit: for the columns) that has eltype `Union{}` and zero elements. You could e.g. create an `struct EmptyVector <: AbstractVector{Union{}}; end` (or something similar) for this, which will probably behave nicely under `vcat` and so-on.

andyferris:
> Insead of AccessStyle traits

I feel that maybe we shouldn't remove this completely... an array's `IndexStyle` isn't about what interface it supports (all arrays support both linear and Cartesian indexing) but about which is faster. Is there some way to reflect upon this in this new design?

> the union of two interfaces in Base: Iterable and what I'm calling PropertyAccessible

Yes! I've been wondering about this for a little while, too. In fact, I was speculating to myself if we could simplify the entire thing and say this: a "table" supports both: you can iterate things that are property accessible (rows), AND you can get properties which are iterable (columns), and these are equivalent views of the content. Anyway, that's what I went with for the new TypedTables.

I realize this intersection is a much narrower scope than intended here. I just wonder if we should imagine complementary sets of interfaces (iterable rows, property-accessible columns, and both) and call them by three different names? Does that make sense?

src/Tables.jl
nalimilan: AFAIK the eltype of the iterator shouldn't be `Union{}`, it should be a `NamedTuple` with the specified fields; they just hold empty vectors.

andyferris: Sorry, I probably wasn't clear, I mean the columns should be like a `Vector{Union{}}()` instead of a `Vector{Missing}()`.

But it's also ever-so-slightly wasteful to allocate `Vector` here for containers you know are empty (this is only important in the case you have many, many small or empty tables... sorry, I'm the StaticArrays guy so I tend to think of these cases and try to minimize overhead for small containers...)

quinnj:
Ok, I just pushed a commit that further refines the unknown schema case according to feedback from review and my comments the other day. In particular:
* tables implement `Tables.rows` and/or `Tables.columns` as before
* tables should implement `Tables.schema(rows_or_columns)` on the result of `Tables.rows` or `Tables.columns` to return either
  * `Tables.Schema` object
  * `nothing`, indicating unknown schema (column names and types)

The fallbacks are now in a separate `fallbacks.jl` file, and are updated to accurately reflect the unknown schema case, as well as the reference implementations. I've also updated the [CSV.write implementation](https://github.com/JuliaData/CSV.jl/blob/9813fadc954e3a6cdf7c52baf7de34b9d0c1af7a/src/write.jl#L119) and [DataFrames implementation](https://github.com/JuliaData/DataFrames.jl/blob/8e2d0df79014dc797e9427d8147f5b3edbe5f4fe/src/other/tables.jl) to reflect these changes.

@andyferris, in response to removing `AccessStyle`, I guess I wonder where it would really be used? For sinks (at least in my implementing so far), always have a "preferred" access style, either they want rows or columns, so they just call what is most natural for them and run w/ it.

src/fallbacks.jl
```diff
@@ -0,0 +1,121 @@
...
+@inline function allocatecolumns(::Schema{names, types}, len) where {names, types}
+    if @generated
+        vals = Tuple(:(allocatecolumn($(fieldtype(types, i)), len)) for i = 1:fieldcount(types))
+        return :(NamedTuple{names}(($(vals...),)))
+    else
+        return NamedTuple{names}(Tuple(allocatecolumn(fieldtype(types, i), len) for i = 1:fieldcount(types)))
+    end
+end
+
+haslength(x) = x === Base.HasLength() || x === Base.HasShape{1}()
```
nalimilan: Is this actually needed? For example, `Base.IteratorSize(Matrix) isa Base.HasShape{2}`, yet you can call `length` and use linear indexing just like with a vector.

quinnj: We could expand the check to be `x === Base.HasLength() || x isa Base.HasShape`

src/fallbacks.jl
```diff
@@ -0,0 +1,121 @@
...
+end
+
+function _buildcolumns(rowitr, row, st, sch, L, columns, rownbr, len, updated)
+    while true
+        eachcolumn(add_or_widen!, sch, row, L, columns, rownbr, len, updated)
+        rownbr += 1
+        state = iterate(rowitr, st)
+        state === nothing && break
+        row, st = state
+        columns !== updated[] && return _buildcolumns(rowitr, row, st, sch, L, updated[], rownbr, len, updated)
```
quinnj, line 107: @piever, this is similar to the idea you proposed for making the iteration loop more type-stable. Here I'm passing in a `Ref{Any}(columns)` when we set/push each row value, and if a column needs to widen, it will update the ref w/ `merge(updated[], NamedTuple{(nm,)}((new_column,))`. A tricky thing that makes this work is the fact that all of the non-widened columns are identical objects between the `columns` and `updated[]` arguments; i.e. the `merge` call makes a new namedtuple that points to all the original columns except the one new widened column. And since we only update each column once per row, we can just check at the end of a row loop if the `updated[]` changed and re-dispatch on that.

This approach seems to perform really well on some larger datasets I tested, comparing the type-stable `columntable(f)` vs. `Tables.buildcolumns(nothing, f)`.

andyferris:
> in response to removing `AccessStyle`, I guess I wonder where it would really be used?

That's a very interesting question.

There are containers that already support (or could reasonably be expected to support) both access styles, with different speeds - things like `DataFrame`s, *TypedTables* `Table` and `FlexTable`, etc - I could even create a view of a column of a CSV file with relatively decent performance given you've already identified the offsets for all the line beginnings (and it would be really cool if you implemented `getproperty` to do exactly this). If I've got a (generic) function which happens to transform a `Table` to a `Table`, which style should it pick, when both access predicates return `true`? Depending on the operation, it might be faster to deal with rows or with columns from a `Table`, but in this case generally operations on columns might be a better default, similar to how linear indexing is generally faster for arrays that support it (but linear indexing is not suitable/ergonomic for all operations - we don't use it for multidimensional `broadcast`ing, for instance).

src/fallbacks.jl
quinnj: Well, would you look at [that](https://github.com/JuliaLang/julia/blob/3fbdc9bbc97e4e5b1d682883ad9d14be5d65c862/base/generator.jl#L98)

src/fallbacks.jl
```diff
@@ -0,0 +1,121 @@
...
+# when Tables.schema(x) === nothing
+function buildcolumns(::Nothing, rowitr::T) where {T}
+    state = iterate(rowitr)
+    state === nothing && return NamedTuple()
+    row::eltype(rowitr), st = state
+    names = propertynames(row)
+    L = Base.IteratorSize(T)
+    len = haslength(L) ? length(rowitr) : 0
+    sch = Schema(names, nothing)
+    columns = NamedTuple{names}(Tuple(Union{}[] for _ = 1:length(names)))
```
quinnj: @piever @davidanthoff @nalimilan, I've thought of one more thing we should probably discuss here before merging. In this line, I'm defaulting the arrays to `Union{}[]`, which will then widen to actual encountered types while iterating. Reading back over [this monster thread](https://discourse.julialang.org/t/missing-data-and-namedtuple-compatibility/8136/80) however, I remembered that people have this paralyzing fear of this "collect-style" table building losing their original table's missingness. For example, with the current implementation here, we have:
```julia
julia> f = CSV.File(joinpath(dir, "test_basic.csv"))
CSV.File(/Users/jacobquinn/.julia/dev/CSV/test/testfiles/test_basic.csv, rows=3):
Tables.Schema:
 :col1  Union{Missing, Int64}
 :col2  Union{Missing, Int64}
 :col3  Union{Missing, Int64}

julia> f |> columntable
(col1 = Union{Missing, Int64}[1, 4, 7], col2 = Union{Missing, Int64}[2, 5, 8], col3 = Union{Missing, Int64}[3, 6, 9])

julia> Tables.buildcolumns(nothing, f)
(col1 = [1, 4, 7], col2 = [2, 5, 8], col3 = [3, 6, 9])
```
So the collect-strategy drops the missingness, though it's notable that in the "inferred" case (or rather when the source knows its schema, which should be the vast majority of cases), the sink can pass that schema on just fine. Now, I think we have a few different options of things to do here:
* Provide at least a keyword argument, so that people could do, essentially, `f |> x->DataFrame(x; allowmissing=:all)`, (to use the CSV.jl keyword arg `allowmissing`, which can be `:auto`, `:none`, or `:all`). That would essentially control whether we do `Union{}[]` or `Missing[]` columns by default when collecting.
* What should be the default? `Union{}[]`? Or `Missing[]`; I'm inclined to say `Union{}[]`, because I think people over-worry about this case, whereas in practical cases, it will be very much controlled by individual sinks. Sure a DataFrame might lose missingness, but that's a very easy case to handle (i.e. vcat will auto-promote columns for you). For a CSV.write operation, it couldn't care less what the column types are in the first place! Similarly for a database sink, the table will have been pre-defined for columns to allow null values or not, which is completely independent of uninferrable iterators that might come along.

davidanthoff: Is the problem here that there is no equivalent to ``fieldtype`` for properties in base? Really if there was a ``propertytypes``, this wouldn't be an issue at all, right? You would look at the first row you get, you could tell that the type of a given property is say ``Union{T,Missing}`` and you could allocate the appropriate array.

Maybe just define that function, provide a default fallback that uses ``fieldtype``, and tell folks that if they use a type that adds a ``getproperty``, they also need to add a method for ``propertytypes``?

Or alternatively, catch the special case of a ``NamedTuple`` and use ``fieldtype`` to do the same thing, and lobby that base adds a ``propertytypes`` function?

quinnj: I don't think that would really solve the problem though of sometimes losing missingness; if my row happens to be `NamedTuple` and none of the individual iterated NamedTuples had a missing value for a column, the resulting column wouldn't allow missingness, even though that might be desirable. That's why I think it might be useful for a user to at least pass something like `DataFrame(x; defaultmissing=true)` and the columns would all pop out like `Union{T, Missing}`.

davidanthoff: Why not? The field types in the ``NamedTuple`` would be ``Union{T,Missing}``, even if there might never be a ``missing`` value, and if in the sink you could query the field type, you can pick the right column type. At least for tables that come from Query.jl that should definitely work, as long as the translator from ``DataValue`` makes sure that if a field in the named tuple is of type ``DataValue``, the field in the named tuple that is returned by the iterator from Tables is of type ``Union{T,Missing}``. I don't see why that wouldn't work.

And in general, I think this problem would *only* show up in a projection case a la Query.jl, right?

src/fallbacks.jl
nalimilan: Good trick! How about adding a comment? ;-)

src/fallbacks.jl
nalimilan: Indeed, since `NamedTuple` is a standard parametric type (contrary to `Tuple`), its instances can encode the information about whether a value could have been missing in its parameters. Having a `propertytype` function in Base would make sense.

In the meantime, we could define `propertytype` internally without exporting it, have a method for `NamedTuple` (based on `fieldtype`), and require custom row types to define it.

src/fallbacks.jl
piever: Yes! Definitely a "best of both worlds" implementation. I agree with @nalimilan that this needs comments in the code as it's a really smart strategy.

piever:
Concerning the discussion about "strictifying things", here's my take. For `NamedTuple` iterators of unknown type, it will be inevitable to strictify things at times: `(f(x) for x in rows(df))` with `f` type unstable or non-inferrable for example. I think we should accept this (esp. in the case of `Missing`) and encourage a programming style that's compatible with the type of the output occasionally strictifying depending on runtime properties.

In particular I think we should provide users with a set of functions that make this easier. Instead of `append!(df, rows)` that would fail if somehow `df` became stricter than expected, we should have a version of `buildcolumns` that starts with an already initialized sink, so one would write `df = append_or_widen!(df, rows)` (obviously with a better name). It should be easy to get this function just by splitting:

```julia
@inline function buildcolumns(schema, rowitr::T, iT=nothing) where {T}
    L = Base.IteratorSize(T)
    len = Base.haslength(L) ? length(rowitr) : 0
    nt = allocatecolumns(schema, len)
    for (i, row) in enumerate(rowitr)
        eachcolumn(add!, schema, row, L, nt, i)
    end
    return nt
end
```

into the part which preallocates `nt` and the part that appends to it .

We should also have an equivalent of  `collect_columns_flattened` (see [here](https://github.com/JuliaComputing/IndexedTables.jl/blob/master/src/collect.jl#L82)) to handle the `by` case where different groups could return columns with different levels of strictification and we want to concatenate them all together (to be honest, it maybe enough to optimize `columns(Iterators.flatten(rows))`.

The case of `setindex!` is trickier but setting a value to be `Missing` seems like a strange thing to do anyway.

quinnj:
Ah, great idea @piever on splitting `Tables.buildcolumns` into the pre-allocation part & the `append_or_widen!` part. Let me play around w/ that and also find a way to incorporate #12, which is related.

src/fallbacks.jl
piever: I commented more extensively below (in particular I do think there will always be cases where strictification happens), but I agree here that the result of `columns([NamedTuple{(:a, :b), Tuple{Int, Union{Int, Missing}}}((1, 2))])` should allow missing values in the second column and this could be done via `propertytype`

Issue #13, opened 2018-09-05
Nesting


piever:
Sorry for opening so many issues, but I am really hoping that this package can be feature rich enough to accomodate all needs from IndexedTables.jl and StructArrays.jl. One question that I was thinking about is nesting. 

Meaning, if I have "nested rows" (say `[(a = (a=1, c=33), b = (c = 1, e = 14))]`), then ideally one would like to have some version of `columns` that returns `(a = (a=[1], c=[33]), b = (c = [1], e = [14]))`.

Would that be possible with Tables currently?

quinnj:
Currently we have:
```julia
julia> t = [(a = (a=1, c=33), b = (c = 1, e = 14))]
1-element Array{NamedTuple{(:a, :b),Tuple{NamedTuple{(:a, :c),Tuple{Int64,Int64}},NamedTuple{(:c, :e),Tuple{Int64,Int64}}}},1}:
 (a = (a = 1, c = 33), b = (c = 1, e = 14))

julia> ct = t |> columntable
(a = NamedTuple{(:a, :c),Tuple{Int64,Int64}}[(a = 1, c = 33)], b = NamedTuple{(:c, :e),Tuple{Int64,Int64}}[(c = 1, e = 14)])

julia> ct.a
1-element Array{NamedTuple{(:a, :c),Tuple{Int64,Int64}},1}:
 (a = 1, c = 33)

julia> ct.a |> columntable
(a = [1], c = [33])
```
So, I'm not quite sure what the question is: if we should try to do this nested unwrapping automatically? Or is the fact that this "works" good enough? Like, in your application, if you knew there would be nested rows, you'd just make sure to pivot to columns, then further pivot the resulting top-level keys.

Let me know if that makes sense and feel free to correct my understanding of what you're after here.

piever:
My evil plan is to use the machinery here to replace:
1. the current implementation of `collect_columns` in IndexedTables (it scales poorly with many "missable" columns, whereas this one here should be fine I think)
2. a collection mechanism for StructArrays (old one is [here](https://github.com/simonster/StructsOfArrays.jl/blob/master/src/StructsOfArrays.jl#L11)).

[...]

PR #17, opened 2018-09-20
WIP: support unwrapping

piever:
This is an attempt at addressing #13. It is still WIP (I've only done the case with known schema, haven't added tests and have probably forgotten to added the `unwrap` keyword in a few places) and lacks tests but I hope is good enough to give an idea. It works as follows:

the user passes a `unwrap` optional argument (defaults to `t -> false`) that specifies what types should be unwrapped. So the default behavior is unchanged:

```julia
julia> x = [(a=(b=1,), c=2)]
1-element Array{NamedTuple{(:a, :c),Tuple{NamedTuple{(:b,),Tuple{Int64}},Int64}},1}:
 (a = (b = 1,), c = 2)

julia> Tables.columns(x)
(a = NamedTuple{(:b,),Tuple{Int64}}[(b = 1,)], c = [2])

julia> unwrap(x) = false
unwrap (generic function with 1 method)

julia> unwrap(::Type{<:NamedTuple}) = true
unwrap (generic function with 2 methods)

julia> Tables.columns(x, unwrap=unwrap)
(a = (b = [1],), c = [2])
```

From experimentation in StructArrays the compiler is quite good here so it figures out the value of `unwrap(T)` in advance (at least in general I couldn't find traces of `unwrap` in `@code_llvm` and results were inferred correctly).

In #13 I was wrong: there's actually no need to add extra `setindex!` or `push!` method provided one is a bit careful in the definition of `add!`

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -44,25 +44,32 @@ haslength(L) = L isa Union{Base.HasShape, Base.HasLength}
     
     Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
 """
-allocatecolumn(T, len) = Vector{T}(undef, len)
+function allocatecolumn(::Type{T}, len; unwrap = t -> false) where {T}
+    if unwrap(T)
+        names = fieldnames(T)
+        types = map(t -> fieldtype(T, t), names)
+        allocatecolumns(Tables.Schema(names, types), len; unwrap = unwrap)
+    else
+        Vector{T}(undef, len)
+    end
+end
 
-@inline function allocatecolumns(::Schema{names, types}, len) where {names, types}
+@inline function allocatecolumns(::Schema{names, types}, len; unwrap = t -> false) where {names, types}
     if @generated
-        vals = Tuple(:(allocatecolumn($(fieldtype(types, i)), len)) for i = 1:fieldcount(types))
+        vals = Tuple(:(allocatecolumn($(fieldtype(types, i)), len; unwrap = unwrap)) for i = 1:fieldcount(types))
         return :(NamedTuple{names}(($(vals...),)))
     else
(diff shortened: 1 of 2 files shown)


src/fallbacks.jl
```diff
@@ -44,25 +44,32 @@ haslength(L) = L isa Union{Base.HasShape, Base.HasLength}
     
     Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
 """
-allocatecolumn(T, len) = Vector{T}(undef, len)
+function allocatecolumn(::Type{T}, len; unwrap = t -> false) where {T}
```
piever: mhm, actually if this is supposed to be overloaded (say to get a `DataValueArray` or things like that), I should add an extra step in between and not touch this one. So `allocatecolumns` should call some `_allocatecolumn` that decides whether calling `allocatecolumn` or to call back `allocatecolumns`.

piever:
I got stuck in the implementation of the type unstable case (in the IndexedTables scenario I think I understand how to do it in full generality but here it doesn't play particularly nicely with the `updated` trick. I was wondering whether this is the wrong approach (as the code in Tables is quite nice and I woulnd't want to mess it up for this one usecase). Still, I hope this WIP PR is enough to give an idea of the strategy I had in mind.

I'm curious whether this is best done at the level of the element being iterated. For example I could define a:
```
struct Unwrapped{F<: Function, T}
   unwrap::F
   itr::T
end
```

whose `propertynames` are the unwrapped propertynames of `el` (with some unique names). For example in the case of `(a=(b=1, c=2), d=3)` the unwrapped propertynames would be `(b, c, d)` (some care is needed to avoid repetition) and everything would proceed as usual, but somehow the information of the row type is kept so that one can remap the collected columns in the end to turn `(b=vb, c=vc, d = vd)` into `(a=(b=vb, c=vc), d=vd)`. Does that make sense?

PR #18, opened 2018-09-22
Switch to use promote_type instead of promote_typejoin

quinnj:
cc: @nalimilan


nalimilan (approved):
Looks good, thanks! It would be nice to test this too.

quinnj:
Good call; we had a test for this, but it wasn't quite specific enough (just used `isequal` to compare `Real[1, 2.0, 3]` and `[1.0, 2.0, 3.0]`). I've adjusted the tests.

PR #23, opened 2018-10-01
Fix indexing

JackDunnNZ:


quinnj:
Thanks for the PR! Did you have cases where you were hitting the non-generated code paths here? I'd be interested if so.

JackDunnNZ:
Nothing real, only when I ctrl-c'd in the middle of reading and it fell back to the non-generated path. Figured I may as well fix it :)

quinnj:
Cool; thanks!

PR #24, opened 2018-10-02
Fix datavaluetype function

davidanthoff:
Fixes https://github.com/queryverse/Query.jl/issues/205.

It also seems like the tests for the Queryverse integration don't run in a normal test run? Could that be changed? Not running them seems quite brittle...


quinnj (approved):
Thanks for the PR!

PR #31, opened 2018-10-19
Experiment with partial support for Generators as a table

quinnj:


andyferris:
I like the simplicity!

I guess it would be nice to have a way of converting a generator of rows into a `columnaccessible` collection.

src/namedtuples.jl
```diff
@@ -117,3 +117,8 @@ function ctappend(ct1::NamedTuple{N1, T1}, ct2::NamedTuple{N2, T2}) where {N1, T
 end
 
 columntable(ct::ColumnTable, itr) = ctappend(ct, columntable(itr))
+
+# partial support for Generators
+rowaccess(::Type{<:Base.Generator}) = true
```
andyferris, line 122: Seem to have missed `istable`

quinnj: It was intentional because of @davidanthoff's concern that `Tables.istable(::Generator)` might not be true. Even this definition isn't strictly true for all Generators (i.e. that they iterate something property-accessible). I just had a thought though that maybe we should allow types to define `Tables.istable` and `Tables.rowaccess` on _instances_ of their types and not just on the types themselves (as is currently required), with a fallback that would call the same method on the type of the object. That would potentially allow us to define `istable` and `rowaccess` on an _instance_ of a Generator that actually satisfied the interface.

davidanthoff: Could one implement ``istable`` on an instance, though? For generators, one would have to start iterating in ``istable``, right?

andyferris: I guess we have the same problem with a `Vector{Any}` that is filled with `NamedTuple`s - this can be a table, but we have to either assert that all Vectors are row tables or else check the instance - and then we still don't want additional errors for tables with no rows! Tricky, tricky...

quinnj:
Superseded by #60

PR #36, opened 2018-10-23
fix module name of Schema to Tables in docs

tpapp:
Editor also removed some dangling spaces, left those in the commit.


quinnj (approved):
Thanks!

PR #38, opened 2018-10-24
Allow collection of higher-dimensional containers by buildcolumns

### LINKED ISSUE #37: Can we make `ColumnTable` allow arbitrary dimenensions?
It seems to me that things like `(a = zeros(3, 3), b = fill(false, 3, 3))` would satisfy the requirements of a length-9 column table. For example, `TypedTables.Table` can be an `AbstractArray` of `NamedTuple`s of arbitrary dimensionality (matching the columns themselves). Do you think we could slightly expand the `ColumnTable` type?

andyferris:
First attempt at to fix #37 

Combined with #31, I can now build `TypedTables.Table`s out of `Generators` involving more than one input. E.g `Table(merge(row1, row2) for row1 in table1, row2 in table2)` (and I particularly enjoyed the inner join `Table(merge(row1, row2) for row1 in table1, row2 in table2 if predicate(row1, row2))`!).

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -56,7 +56,7 @@ allocatecolumn(T, len) = Vector{T}(undef, len)
 end
 
 # add! will push! or setindex! a value depending on if the row-iterator HasLength or not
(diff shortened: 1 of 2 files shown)


quinnj (approved):
These changes look fine to me.

Issue #32, opened 2018-10-20
Allow table type to specify constructor as part of the interface


piever:
I think this is not part of the interface as of now, but it'd be nice to have a way to know, for a given type that respects the table interface, how to build a table of that type.

The use case is that we may have some table operation that we want to work on any table, like

```julia
function mygroupby(s::T, args...) where {T}
  ....
  return t
end
```

If `t` is table-like, one may want to recreate a table of type `T` from `t` (for example, build back a `DataFrame`).

quinnj:
What should we call this? A few ideas that come to mind:
* `Tables.sinkfunction`
* `Tables.sink`
* `Tables.constructor`

quinnj:
This also gets a little murky when considering, for example, CSV.jl. The `CSV.File` type implements the Tables.jl source interface, but `CSV.write` takes any Tables.jl source and writes out a csv file. I guess we need to clarify what exactly we're after here and the expected use-cases: taking Tables.jl sources and returning some kind of in-memory structure? Just getting the sink function to a corresponding source type?

piever:
I initially thought as `Tables.constructor` as a name, so if that also came to your mind independently I guess it's a good solution.

The purpose is to be able design "table-implementation independent" packages that also return tables. For example, I'm creating a set of widgets to do manipulations on tabular structures from a GUI. I'd like to support anything that implements the `Tables` interface. I assume that if the user run some operations from the GUI on a `DataFrame`, even though under the hood I'm not using anything from DataFrames but just the `Tables` machinery, I expect they'd like my GUI to give them back a `DataFrame`. Similarly for `IndexedTable`.

I would define both a trait `Tables.hasconstructor` to specify whether there is a constructor that makes sense (which in the case of `CSV.File` I guess there isn't) and a `Tables.constructor` function that returns the specific constructor. In this way for example in the case of the GUI, if one gets as input a `DataFrame`, or `IndexedTable` or `TypedTable` or `Array{NamedTuple}` or what not the GUI would return the corresponding structure. If one instead starts with some structure for which this is impossible or ill-defined (like a csv file for example), I think it makes sense to default to return a lazy table.

piever:
> Just getting the sink function to a corresponding source type?

This would be enough: return a function that can ingest any table and return something of the source type (as mentioned above, I believe that for user unaware of Tables it can be a usability improvement in many cases to get something of the type they're familiar with - e.g. `DataFrames` - rather than some abstract table object).

[...]

PR #41, opened 2018-11-03
add materializer function

joshday:
Ref: https://github.com/JuliaData/Tables.jl/issues/32

This PR adds ~~a single line~~ a few lines to the source:

```
materializer(x) = columntable

materializer(x::RowTable) = rowtable
```

along with instructions in the README of what `materializer` is designed to do.

I would like to add some more tests, but thought I'd open this up for comments.

- [x] tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,9 +87,9 @@ Obviously every table type is different, but via a combination of `Tables.rows` 
 
 ## Functions that input and output tables:
 
-For functions that input a table, perform some calculation, and output a new table, we need a way of constructing the preferred output table given the input.  For this purpose, the `Tables.materializer(table)` function returns the the constructor for the preferred in-memory structure of the input table.  By default, `Tables.materializer(table)` returns `identity`.
+For functions that input a table, perform some calculation, and output a new table, we need a way of constructing the preferred output table given the input.  For this purpose, `Tables.materializer(table)` returns the preferred sink function for a table (`Tables.columntable`, which creates a named tuple of AbstractVectors, is the default).  
 
-Note that an in-memory table with a properly defined "sink" constructor can reconstruct itself with either of the following:
+Note that an in-memory table with a properly defined "sink" function can reconstruct itself with the following:
 
 ```julia
 materializer(table)(columns(table)) 
--- a/README.md
+++ b/README.md
@@ -108,8 +108,11 @@ function select(table, cols::Symbol...)
 end
 
 # Example of selecting columns from a columntable
-nt = (x=1:100, y=rand(100), z=randn(100))
-select(nt, :x)
-select(nt, :x, :z)
+tbl = (x=1:100, y=rand(100), z=randn(100))
+select(tbl, :x)
+select(tbl, :x, :z)
+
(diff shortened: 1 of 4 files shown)


src/Tables.jl
```diff
@@ -123,6 +123,7 @@ rowaccess(::Type{T}) where {T} = false
 columnaccess(x::T) where {T} = columnaccess(T)
 columnaccess(::Type{T}) where {T} = false
 schema(x) = nothing
+materializer(x) = identity
```
nalimilan: How useful is this default? This means we'll return a row- or a column-table depending on the internal implementation. Wouldn't it be better to throw an error, or always use the same kind of representation?

README.md
```diff
@@ -76,3 +84,35 @@ end
 ```
 
 Obviously every table type is different, but via a combination of `Tables.rows` and `Tables.columns` each table type should be able to construct an instance of itself.
+
+## Functions that input and output tables:
+
+For functions that input a table, perform some calculation, and output a new table, we need a way of constructing the preferred output table given the input.  For this purpose, the `Tables.materializer(table)` function returns the the constructor for the preferred in-memory structure of the input table.  By default, `Tables.materializer(table)` returns `identity`.
```
nalimilan: What guarantees do we want to provide? Should the table be mutable? It would be nice to be explicit about this.

If we want the table to be mutable, I still think `similar` would be a good function for this.

src/Tables.jl
joshday: My thinking is that it wouldn't break anything, and then the user can decide what to do with the row or column table that's returned, rather than possibly making a bunch of copies to go from row to column or vice versa.  I'd be okay with throwing an error or choosing `rowtable`/`columntable`

README.md
joshday: I think `materializer` is different enough from `similar` that it would cause confusion.  `materializer(tbl)` does not return a table; it returns a function.

E.g.

```
materializer(x::ColumnTable) = columntable

materializer(x::DataFrame) = DataFrame
```

So that `materializer(tbl)(other_tbl)` creates the preferred table type of `tbl`.

nalimilan: Yes, my suggestion was to write `similar(table, columns)`.  But indeed that would be weird. It would have to be something like `copy!(similar(table, schema), columns)` to be consistent, but it's quite convoluted.

joshday:
I've changed the default definition to `materializer(x) = columntable`.  It's easy enough to add that I think most tables won't rely on the default, but @nalimilan made the good point that it shouldn't rely on the internal implementation of table operations.  I also think an error would be too severe.

README.md
joshday: Ah, okay.  The assumption of a mutable table may be too strong.  Since sink functions are already encouraged by the interface, I'm partial to this current approach of "give me the sink function for this table".

kleinschmidt:
Why not use the `rowaccess`/`columnaccess` traits to choose the default behavior?  Say, use `columntable` if `columnaccess`, and `rowtable` otherwise?

Also I thought at some point there was a "preferred access" trait but I don't see that in the README so maybe it got dropped.

quinnj (approved):
Looks great! Thanks for doing this!

PR #44, opened 2018-11-12
Shim IteratorSize for DataValueRowIterator

### LINKED ISSUE #43: DataValueRowIterator incorrectly implements the IteratorSize trait
If the source of a ``DataValueRowIterator `` returns ``HasShape()`` for ``IteratorSize(x)``, then currently ``DataValueRowIterator`` just passes that value through, but then does not implement ``size``.

I think the solution would be to either properly implement ``size``, or to not just pass the value of ``IteratorSize(x)`` through, but e.g. return ``HasLength()`` if the source implements a size with 1 dimension or something like that.

This showed up as a bit in https://github.com/fredo-dedup/VegaLite.jl/issues/129.

davidanthoff:
Fixes #43, I think...

### DIFF
--- a/src/datavalues.jl
+++ b/src/datavalues.jl
@@ -29,8 +29,11 @@ function datavaluerows(x)
     return DataValueRowIterator(datavaluetype(Tables.schema(r)), r)
 end
 
+_iteratorsize(x) = x
+_iteratorsize(::Base.HasShape{1}) = Base.HasLength()
+
 Base.eltype(rows::DataValueRowIterator{NT, S}) where {NT, S} = NT
-Base.IteratorSize(::Type{DataValueRowIterator{NT, S}}) where {NT, S} = Base.IteratorSize(S)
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Seems like a fine solution to me. Thanks!

src/datavalues.jl
```diff
@@ -29,8 +29,11 @@ function datavaluerows(x)
     return DataValueRowIterator(datavaluetype(Tables.schema(r)), r)
 end
 
+_iteratorsize(x) = x
+_iteratorsize(::Base.HasShape{1}) = Base.HasLength()
```
nalimilan, line 33: I don't get the reasoning here. Why should we treat `HasShape{2}` differently from `HasShape{1}`?

davidanthoff: I don’t believe this will ever kick in for anything other then ``Vector``. I’m not sure, though, this is all getting very complex...

davidanthoff:
@quinnj Could you also tag a new release?

PR #45, opened 2018-11-14
Allow any TableTraits source to satisfy the Tables interface

quinnj:


quinnj:
Ok, I think we should go ahead with this; the Tables.jl inteface aims for maximum convenience and interoperability, and the added dependencies aren't too heavy. This will allow package developers to just Tables.jl and get _everything_ else for free, without needing to worry about extra integrations.

src/Tables.jl
```diff
@@ -116,7 +118,8 @@ Obviously every table type is different, but via a combination of `Tables.rows`
 abstract type Table end
 
 # default definitions
-istable(x::T) where {T} = istable(T)
+istable(x::T) where {T} = istable(T) || TableTraits.isiterabletable(x) === true ||
+    TableTraits.isiterabletable(x) === missing
```
davidanthoff, line 122: Is this a good idea? That essentially means that any consumer of ``istable`` must now be prepared that even if ``istable`` returns ``true``, that it actually is not a table...

src/fallbacks.jl
```diff
@@ -146,7 +150,17 @@ end
     if rowaccess(T)
         r = rows(x)
         return buildcolumns(schema(r), r)
+    elseif TableTraits.supports_get_columns_copy_using_missing(x)
+        return TableTraits.get_columns_copy_using_missing(x)
```
davidanthoff, line 154: I'm also not sure this is in general a good idea. What are the copy semantics of ``columns``? I.e. as a client, will I get a set of vectors that I "own", or is it a view? If it is the former, this makes sense. But it the semantics are the later, I think this would not be a good implementation. From my reading of the README I would think that the semantics of ``columns`` is that of a view, namely an iterator, but then it seems to be used in quite a few places in different ways and I'm just not sure. I think this might be a another case where [TableTraits.jl](https://github.com/queryverse/TableTraits.jl) makes are more fine grained distinction between different interfaces (copy and view column interfaces) and now putting them behind one combined column interface seems not like a good idea.

PR #50, opened 2018-12-07
make rowiterator an abstractvector

piever:
This I think simplifies the code and adds feature as we can use the `AbstractVector` machinery to iterate and get an element by index / slice the table etc...

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -48,6 +48,7 @@ struct RowIterator{T, U} <: AbstractVector{U}
 end
 Base.size(x::RowIterator) = (x.len,)
 Base.getindex(x::RowIterator, i::Int) = ColumnsRow(x.columns, i)
+Base.IndexStyle(::Type{<:RowIterator}) = IndexLinear()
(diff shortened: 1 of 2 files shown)


nalimilan:
Makes sense, but you need to define `Base.IndexStyle` to return `IndexLinear` too I think.

quinnj:
Can we do some performance tests to just make sure this doesn't affect anything there?

piever:
Sure, I'll try to find some time today: I should probably specify that this abstractvector prefers linear indexing (as suggested above) and use `@boundscheck` and `@inbounds` to make sure we only do inbounds check when necessary (i.e, not when iterating the vector).

EDIT: actually the `@inbounds` change is maybe best left to a later PR as it may be not 100% trivial

Do you have some standard benchmarks for iteration?

quinnj:
I don't have anything standard, I usually use these lines from CSV to sanity check things (https://github.com/JuliaData/CSV.jl/blob/master/test/files.jl#L56); the pandas csv file is big enough to hit the row-iteration and the `buildcolumns` call hits the no-schema path.

PR #51, opened 2018-12-07
implement isless for ColumnsRow

piever:
This is a possible implementation of `isless` for `ColumnsRow`. Happy to add tests if we are happy with the feature and implementation.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -17,7 +17,7 @@ Base.propertynames(c::ColumnsRow) = propertynames(getfield(c, 1))
 function _isless(c, d, t::Tuple)
     f = t[1]
     a, b = getproperty(c, f), getproperty(d, f)
-    isless(a, b) || isequal(a, b) && _isless(c, d, tail(t))
+    isless(a, b) || isequal(a, b) && _isless(c, d, Base.tail(t))
 end
 
 _isless(c, d, ::Tuple{}) = false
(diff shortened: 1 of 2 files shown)


piever:
UPDATE: I've switched to generated functions as the performance was horrible otherwise (I suspect it may be due to `propertynames` not being constantly inferred but am not sure). It still seems slightly slower than StructArrays (against which I'm benchmarking) but acceptable. I'll try to post detailed benchmarks later.

piever:
Benchmarking, this version seems to perform as well as a `StructArray`:

```julia
julia> using Tables, StructArrays

julia> using BenchmarkTools

julia> N = 10000;

julia> x = rand(1:10, N);

julia> y = rand(1:10, N);

julia> z = rand(1:10, N);

julia> t1 = Tables.rows((x=x, y=y, z=z));

julia> t2 = Tables.rowtable((x=x, y=y, z=z));

julia> t3 = StructArray(x=x, y=y, z=z);

julia> function lessthanfirst(v)
           fr = first(v)
           c = 0
           for row in v
               isless(row, fr) && (c+=1)
           end
           c
       end
lessthanfirst (generic function with 1 method)

julia> @btime lessthanfirst($t1)
  29.045 μs (0 allocations: 0 bytes)
7929

julia> @btime lessthanfirst($t2)
  22.700 μs (0 allocations: 0 bytes)
7929

julia> @btime lessthanfirst($t3)
  27.795 μs (0 allocations: 0 bytes)
7929
```

However this version is the only one where performance survives in the case of missing data (it really feels like the lazy iteration approach of Tables is the correct one with missings / small type instabilities). Now instead of `rand(1:10, N)` I'm using `rand(vcat(1:10, missing), N)` as columns:

```julia
julia> @btime lessthanfirst($t1)
  38.140 μs (0 allocations: 0 bytes)
9050

julia> @btime lessthanfirst($t2)
  3.543 ms (19992 allocations: 621.23 KiB)
9050

julia> @btime lessthanfirst($t3)
  6.901 ms (39986 allocations: 1.52 MiB)
9050
```

quinnj:
Ok, had a chance to pull this locally and play around; looks good to me. For the slow cases, is the issue that we don't have generated versions for NamedTuples? Or for StructArrays you materialize the struct on each iteration, right? So each struct would need an efficient `isless`?

quinnj (approved):

PR #54, opened 2019-01-08
Patch to make 0.13 usable

Nosferican:
Should fix the
```
ERROR: LoadError: ArgumentError: Package Tables does not have TableTraits in its dependencies:
- If you have Tables checked out for development and have
  added TableTraits as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with Tables
```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,4 @@
 julia 0.7
 Requires
+TableTraits
+IteratorInterfaceExtensions


davidanthoff:
Wouldn't it make most sense to just remove the ``Project.toml`` entirely and just have a REQUIRE file? That seems to be the recommended way of doing it until METADATA is migrated to Pkg3?

Nosferican:
Having the Project.toml does not affect the actual release at all since it is currently ignored completely by attobot/METADATA and General uses the script-generated one. It does require to have the correct REQUIRE. The issue in this case is that the CI does use the Project when available over the REQUIRE which in this case backfired and didn't catch the issue. Guess the safest route is to avoid committing the Project.toml and just have every collaborator generate these every time for every branch. While this is a pain for developing it does eliminate the risk of the CI reporting incorrect results. Best approach would be to have the CI behavior changed to use the REQUIRE when available over the Project.toml.

quinnj (approved):

quinnj:
Thanks @Nosferican!

Issue #53, opened 2019-01-08
0.13.0 Release Broken


Nosferican:
The REQUIRE file is missing dependencies in the Project.toml. This makes it that Tables.jl fail to precompile. The patch should fix it.

Nosferican:
With #54 merged, the patch release should be merged ASAP (maybe a manual trigger to propagate to General?, won't have any changes to METADATA). The 0.13.0 should probably be yanked (@StefanKarpinski could you confirm we should yank that release version since it just fails to precompile). Once these steps are done, this issue should be closed.

quinnj:
The new tag v0.1.14 has been merged in METADATA.

Nosferican:
I think we can close this issue now. To be super save we could ask to yank the 0.1.13 release on General, but since it was less than a day as the latest release it should hopefully be fine regardless.

PR #55, opened 2019-01-08
CI fixes

davidanthoff:
This removes ``Project.toml`` to prevent things like https://github.com/JuliaData/Tables.jl/issues/53 happening again.

@Nosferican  mentioned in https://github.com/JuliaData/Tables.jl/pull/54 that removing ``Project.toml`` would imply that every developer would have to generate their own ``Project.toml`` files:

> Guess the safest route is to avoid committing the Project.toml and just have every collaborator generate these every time for every branch. While this is a pain for developing it does eliminate the risk of the CI reporting incorrect results.

I don't actually understand that point :) I've been developing lots and lots of packages since julia 1.0, none of them have a ``Project.toml``, and I never generate any, and everything works just fine. Am I missing something?

This PR also enables the full sets of platforms in travis and appveyor, I'm actually surprised they weren't enabled in the first place. Plus julia 1.1 support.


Nosferican:
You are probably adding all the packages to the global/default environment and since the environment stack you don't have to maintain the Project.toml. That is the 0.6 style and should be fine, but lacks the check locally that all the dependencies are working as intended (or the ability to have different versions of packages simultaneously as in one per environment). CI should catch the issues. Did you verify CI likes the 1.1 version? I know the branch had a RC version, but don't know if they have tagged it for CI purposes or if it gets updated (or if it is available on Win/Mac). If CI is not covering 0.7, maybe we should up the REQUIRE to julia 1.

davidanthoff:
Yes, the 1.1 version tag seems to work ever since the RC1 was published. I assume that it will automatically get the released version once that is done.

I agree on the 0.7: either it should be covered by tests or REQUIRE should be bumped. I've for now enabled tests on 0.7, that seems the less intrusive change.

> That is the 0.6 style and should be fine, but lacks the check locally that all the dependencies are working as intended (or the ability to have different versions of packages simultaneously as in one per environment).

Ah, so you are actually activating the package env when you work on it? Interesting... I guess I still don't fully understand why... Wouldn't a more realistic story be to just have a shared env where the package is deved? But in any case, I think we agree that whatever downsides the lack of a ``Project.toml`` might have, is worth having if it prevents things like #53 :)

quinnj:
This feels like a step backward; the future is Project.toml, not REQUIRES, so I'd rather not go back and just be more careful with things like adding new dependencies (not that there are remotely any plans to do something similar soon).

PR #60, opened 2019-01-30
Rework the istable and Tables.rows definitions to rely on inputs bein…

quinnj:
…g iterable instead of the TableTraits interface. We also refactor DataValueUnwrapper to be more generically an IeteratorWrapper (which still unwraps DataValues), but also throws an informative error if the iterator doesn't iterate a valid Row object (something property-accessible). This cleans up some corner cases where certain iterables turned into weird or invalid tables.


quinnj:
@nalimilan and @andyferris, I'd appreciate your eyes on this if you wouldn't mind taking a look. @andyferris , probably of particular interest to you, the changes here allow Generators of NamedTuples (or any property-accessible object) to be valid Tables. In the case an iterable input *doesn't* iterate valid property-accessible objects, an informative error is thrown (in iteratorwrapper.jl).

nalimilan:
I'm not sure I'm qualified to review this precisely. Is there any part you'd like me to comment on? Or can you describe the general strategy here if that's what you'd like my input on?

src/Tables.jl
```diff
@@ -118,8 +118,7 @@ Obviously every table type is different, but via a combination of `Tables.rows`
 abstract type Table end
 
 # default definitions
-istable(x::T) where {T} = istable(T) || TableTraits.isiterabletable(x) === true ||
-    TableTraits.isiterabletable(x) === missing
+istable(x::T) where {T} = istable(T) || TableTraits.isiterabletable(x) === true || Base.isiterable(T)
```
davidanthoff, line 121: Is the ``Base.isiterable(T)`` part really a good idea? At that point ``istable`` returns ``true`` for pretty much anything, and for lots of things where it is entirely clear that they are *not* a table. I guess I don't really understand what purpose ``istable`` has at that point...

davidanthoff (commented):
I made one narrow comment about ``istable``, but beyond that I no longer understand what is going on between all the ``@require`` clauses, fallback definitions etc., so I can't really review this.

Are the queryverse tests actually running these days? They all seem behind an ``if`` statement, and it is not clear to me whether they are running during CI or not. I also can't find coverage information anywhere, so difficult to tell.

quinnj:
Most of the info is in the original post/commit message; essentially, if something is iterable, we'll try to treat it as an `Tables.istable`, but now in the `IteratorWrapper` (which implements the tables interface for any iterator), if the elements the iterator produces don't satisfy the `Row` interface, we throw an informative error. So the progression of something like a Generator of Ints has been:
* Originally, would throw an `ArgumentError` that the input wasn't a table
* In the current release, it tries to treat it as a table, but runs into a cryptic error or you could weirdly shaped tables
* With this change: you can an informative error that the iterator input doesn't produce valid elements

I've been sketching out this change for a while, so I feel pretty good about it. I'll merge now, since I think it's a strict improvement over the current situation where we're too liberal in what we consider a table and you end up getting weird tables or cryptic errors.

nalimilan:
OK. But then why add `Base.isiterable(T)` as @davidanthoff noted above? Was it already used somehow?

andyferris:
Hi Jacob - sorry been super busy, I’ll try have a deeper look/think in the future but I am really glad you are iterating (pun may or may not be intended) over this (and naturally this fits my view of tables <-> relations :) ).

src/datavalues.jl
```diff
@@ -24,8 +24,7 @@ DataValueRowIterator(::Type{NT}, x::S) where {NT <: NamedTuple, S} = DataValueRo
 DataValueRowIterator(::Type{Schema{names, types}}, x::S) where {names, types, S} = DataValueRowIterator{datavaluetype(NamedTuple{names, types}), S}(x)
 
 function datavaluerows(x)
-    r = Tables.rows(x)
-    #TODO: add support for unknown schema
+    r = Tables.rowtable(x)
```
davidanthoff, line 27: @quinnj Why was that changed from ``Tables.rows``? If I understand this correctly, it now means that if someone uses Query.jl with a ``DataFrame`` source, it will first allocate a new vector of named tuples, and then pass that on to Query.jl. That seems super inefficient, and really we would want the old behavior here that the named tuples are created per ``iterate`` call and streamed, right?

I think going back to the old behavior might also solve https://github.com/queryverse/Query.jl/issues/240. I'm still trying to understand what is happening there, but if I load that file and then do ``df |> getiterator |> collect``, everything hangs forever. I'm not at the bottom of what is happening there, but on the other hand, having the ``collect`` call in there seems generally undesirable, so maybe that would just solve the bug.

quinnj: Well, it would be better to figure out what exactly is causing the hang. The change here is intentional, because, as shown in the previous code comment, unknown schema weren't supported by this function. By materializing as a Vector of NamedTuples, it handles that case for us. Maybe we can figure out a different solution altogether that doesn't allocate a new vector while still handling unknown schema.

davidanthoff: I think it is really important to go back to the old behavior for the ``DataFrame`` and Query.jl combination. For that scenario there shouldn't be any issue about unknown schemas. At the end of the day, this change here is a major regression for the most common usage scenario for Query.jl: this PR now makes a copy of everything in the DataFrame before any query runs, which really defeats the whole design of Query where multiple query operators get fused and intermediate allocations of results are avoided at all costs when possible.

I understand that we in general want to handle the case of unknown schemas here, but as far as I can tell right now we actually don't have any case in the wild where ``datavaluerows`` could be called with an ``x`` that doesn't have a schema, right? At least right now I can't think of any combination of sinks and sources where such a situation would come up. If that is correct, then I think it is an even stronger argument to undo this change here quickly: I don't think it is the right trade-off in the short term to regress the main usage scenario for a benefit in a scenario that can't actually come up right now.

In terms of a solution: can't we just check ``schema(x)===nothing``  at the beginning of this function here and only then allocate an intermediate array? I don't fully understand

nalimilan: @davidanthoff I'd say make a PR to revert the problematic line until we find a better solution.

(Thouh AFAICT column vectors are not copied when a data frame is passed, they are just put in a named tuple. But I guess that can be costly for the compiler if there are many columns.)

davidanthoff: https://github.com/JuliaData/Tables.jl/pull/67

Issue #59, opened 2019-01-29
Materialising matrix as a table


ablaom:
In our machine learning framework MLJ we have a function that materialises an abstract matrix (or named tuple of vectors corresponding to columns) as a table. Typical use case: a transformer takes a table as input, converts it to a matrix for processing, but then is expected to output a table once again. Perhaps such functionality should lie in Tables.jl? 

Here's our code at the moment:

````julia
"""
    MLJBase.table(cols; prototype=DataFrames)

Convert a named tuple of vectors `cols`, into a table. The table
type returned is the "preferred sink type" for `prototype` (see the
Tables.jl documentation), which is generally the type of `prototype`
itself, or a named tuple of vectors (in which case `cols` itself is
returned).

    MLJBase.table(X; names=nothing, prototype=DataFrames())

Convert an abstract matrix `X` into a table with `names` (a tuple of
symbols) as column names, or with labels `(:x1, :x2, ..., :xn)` where
`n=size(X, 2)`, if `names` is not specified.  Equivalent to `table(cols,
prototype=prototype)` where `cols` is the named tuple of columns of
`X`, with `keys(cols) = names`.

"""
function table(cols::NamedTuple; prototype=DataFrames.DataFrame())
    Tables.istable(prototype) || error("prototype is not tabular.")
    return Tables.materializer(prototype)(cols)
end
function table(X::AbstractMatrix; names=nothing, prototype=DataFrames.DataFrame())
    if names == nothing
        _names = tuple([Symbol(:x, j) for j in 1:size(X, 2)]...)
    else
        _names = names
    end
    cols = NamedTuple{_names}(tuple([X[:,j] for j in 1:size(X, 2)]...))
    return table(cols; prototype=prototype)
end
````  

Incidentally, we also provide convenience functions for selecting rows or columns from a table, returning a table of the "preferred sink type" of the input. Maybe such functionality should also live Tables.jl.

quinnj:
The new `Tables.table` function has been added to wrap a matrix as a `MatrixTable`, which allows using `Tables.rows` and `Tables.columns` to access the underlying matrix data.

Issue #58, opened 2019-01-26
Table to matrix


tlienart:
(apologies if it's a dumb question)

In line with [this discourse thread](https://discourse.julialang.org/t/about-the-convention-for-the-design-matrix-row-as-observation-or-column-as-observation/20080/4) it seems that a good way forward is to have ML algorithms (such as, say, `kmeans`) accept `Tables` instead of "just" `Matrix` or `AbstractMatrix`. 

In some cases, the algorithm may work directly using `rows` or `columns` of the Table but in some case it may be preferable to work with the matrix of values of the table directly (maybe an example would be PCA).

Would it therefore be possible/relevant for `Tables.jl` to implement a function like `Tables.matrix(table)` and possibly have that function take an argument depending on whether the user wants a matrix with rows-as-observation or column-as-observation?

As a suggested possible API I could see something like

```julia
Tables.matrix(table;
  rows=nothing, # to indicate a subset of the rows / nothing for all
  cols=nothing, # same 
  vardim=1) # 1 = row-as-observation, 2 = col-as-observation
```

Edit: here are what I think may be useful building blocks? (with guidance/hints I'm happy to try to give this a shot by the way, though I'm not very familiar with Julia's table universe...)

* have tables implement a `isroworiented` / `iscolumnoriented`
* use something like the conversion code in DataFrames: https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/abstractdataframe.jl#L804_L828
* if it's column oriented and `vardim=2` and the promoted type is `<:Number`, I believe `copy(transpose(matrix(..., vardim=1)))` will be faster than a row iteration, if it's not a `<:Number` and therefore transpose might not work, `permutedims` should.
* similar reasoning for row oriented

nalimilan:
That sounds like a useful feature. Though I think only one argument is needed: `vardim`. Subsetting rows and columns can be achieved using separate functions.

A few remarks:
> * have tables implement a `isroworiented` / `iscolumnoriented`

I don't think this is needed. Tables always have observations as rows and variables as columns. What can change is how the storage is actually done, but the Tables.jl interface ensures we don't have to care about it.

> * use something like the conversion code in DataFrames: https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/abstractdataframe.jl#L804_L828

Yes, this can be an inspiration. But of course this should use `Tables.columns` and `Tables.schema` instead of DataFrames-specific functions.

[...]

PR #61, opened 2019-01-30
Add MatrixTable type to wrap a Matrix and provide Tables.jl interface.

quinnj:
Currently addresses #59, will finish so it also addresses #58.

cc: @ablaom, @tlienart

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -1,6 +1,6 @@
 module Tables
 
-using Requires
+using Requires, LinearAlgebra
 
 using TableTraits, IteratorInterfaceExtensions
 
@@ -187,5 +187,8 @@ include("fallbacks.jl")
 
 # integration with queryverse
 include("query.jl")
+
+# matrix integration
+include("matrix.jl")
 
 end # module
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -1,0 +1,54 @@
+istable(::Type{<:Matrix}) = false
+
+rows(m::T) where {T <: Matrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.astable` for ways to treat a Matrix as a table"))
+columns(m::T) where {T <: Matrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.astable` for ways to treat a Matrix as a table"))
+
+struct MatrixTable{T}
+    names::Vector{Symbol}
+    lookup::Dict{Symbol, Int}
+    matrix::Union{Matrix{T}, Adjoint{T, Matrix{T}}}
+end
+
+istable(::Type{<:MatrixTable}) = true
+names(m::MatrixTable) = getfield(m, 1)
+
+# row interface
+struct MatrixRow{T}
+    row::Int
+    source::MatrixTable{T}
+end
+
+Base.getproperty(m::MatrixRow, ::Type{T}, col::Int, nm::Symbol) where {T} =
+    getfield(getfield(m, 2), 3)[getfield(m, 1), col]
+Base.getproperty(m::MatrixRow, nm::Symbol) =
+    getfield(getfield(m, 2), 3)[getfield(m, 1), getfield(getfield(m, 2), 2)[nm]]
+Base.propertynames(m::MatrixRow) = names(getfield(m, 2))
+
+rowaccess(::Type{<:MatrixTable}) = true
+schema(m::MatrixTable{T}) where {T} = Schema(Tuple(names(m)), NTuple{size(getfield(m, 3), 2), T})
+rows(m::MatrixTable) = m
+Base.eltype(m::MatrixTable{T}) where {T} = MatrixRow{T}
+Base.length(m::MatrixTable) = size(getfield(m, 3), 1)
+
+function Base.iterate(m::MatrixTable, st=1)
+    st > length(m) && return nothing
+    return MatrixRow(st, m), st + 1
+end
+
+# column interface
+columnaccess(::Type{<:MatrixTable}) = true
+columns(m::MatrixTable) = m
+Base.getproperty(m::MatrixTable, ::Type{T}, col::Int, nm::Symbol) where {T} = getfield(m, 3)[:, col]
+Base.getproperty(m::MatrixTable, nm::Symbol) = getfield(m, 3)[:, getfield(m, 2)[nm]]
+
+function astable(m::Union{Adjoint, Matrix}; header::Vector{Symbol}=[Symbol("Column$i") for i = 1:size(m, 2)])
+    length(header) == size(m, 2) || throw(ArgumentError("provided column names `header` length must match number of columns in matrix ($(size(m, 2))"))
+    lookup = Dict(nm=>i for (i, nm) in enumerate(header))
+    return MatrixTable(header, lookup, m)
+end
+
+function asmatrix(table)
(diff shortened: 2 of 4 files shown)


src/matrix.jl
```diff
@@ -0,0 +1,61 @@
+istable(::Type{<:Matrix}) = false
+
+rows(m::T) where {T <: Matrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.astable` for ways to treat a Matrix as a table"))
+columns(m::T) where {T <: Matrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.astable` for ways to treat a Matrix as a table"))
+
+struct MatrixTable{T}
+    names::Vector{Symbol}
+    lookup::Dict{Symbol, Int}
+    matrix::Union{Matrix{T}, Adjoint{T, Matrix{T}}}
```
tlienart: I think here `Transpose` is missing. 

```julia
a = randn(5, 5);
foo(X::Union{Matrix{Float64}, Adjoint{Float64, Matrix{Float64}}}) = norm(X)
foo(a) # ok
foo(a') # ok
foo(transpose(a)) # nok
```

src/Tables.jl
```diff
@@ -1,6 +1,6 @@
 module Tables
 
-using Requires
+using Requires, LinearAlgebra
```
tlienart, line 3: needs to be added in the REQUIRE (seems to be causing CI to fail)

src/matrix.jl
nalimilan: Why not accept any `AbstractMatrix` and parameterize `MatrixTable` on that?

src/matrix.jl
```diff
@@ -0,0 +1,61 @@
...
+end
+
+# column interface
+columnaccess(::Type{<:MatrixTable}) = true
+columns(m::MatrixTable) = m
+Base.getproperty(m::MatrixTable, ::Type{T}, col::Int, nm::Symbol) where {T} = getfield(m, 3)[:, col]
+Base.getproperty(m::MatrixTable, nm::Symbol) = getfield(m, 3)[:, getfield(m, 2)[nm]]
+Base.propertynames(m::MatrixTable) = names(m)
+
+function astable(m::Union{Adjoint, Matrix}; header::Vector{Symbol}=[Symbol("Column$i") for i = 1:size(m, 2)])
```
nalimilan: I don't love these "as*" names. They remind me of the R terminology and sound like they create a parallel `convert` system. Why not just `table`? Or maybe even better make this a convenience constructor for `MatrixTable`.

src/matrix.jl
```diff
@@ -0,0 +1,61 @@
...
+istable(::Type{<:MatrixTable}) = true
+names(m::MatrixTable) = getfield(m, 1)
+
+# row interface
+struct MatrixRow{T}
+    row::Int
+    source::MatrixTable{T}
+end
+
+Base.getproperty(m::MatrixRow, ::Type{T}, col::Int, nm::Symbol) where {T} =
```
nalimilan:
```suggestion
Base.getproperty(m::MatrixRow, ::Type, col::Int, nm::Symbol) =
```

src/matrix.jl
```diff
@@ -0,0 +1,61 @@
...
+    getfield(getfield(m, 2), 3)[getfield(m, 1), col]
```
nalimilan: Wouldn't using symbols instead of indices be as efficient? That would be much more readable.

quinnj:
Thanks @tlienart and @nalimilan for the review and comments; I think I've addressed everything. I also added some tests.

src/matrix.jl
```diff
@@ -0,0 +1,71 @@
+istable(::Type{<:AbstractMatrix}) = false
+
+rows(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))
+columns(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))
+
+struct MatrixTable{T}
```
nalimilan:
```suggestion
struct MatrixTable{T<:AbstractMatrix}
```

src/matrix.jl
nalimilan: This comment is still valid.

src/matrix.jl
```diff
@@ -0,0 +1,71 @@
...
+
+# column interface
+columnaccess(::Type{<:MatrixTable}) = true
+columns(m::MatrixTable) = m
+Base.getproperty(m::MatrixTable, ::Type{T}, col::Int, nm::Symbol) where {T} = getfield(m, 3)[:, col]
+Base.getproperty(m::MatrixTable, nm::Symbol) = getfield(m, 3)[:, getfield(m, 2)[nm]]
+Base.propertynames(m::MatrixTable) = names(m)
+
+"""
+Tables.table(m::AbstractMatrix; header::Vector{Symbol}=...)
```
nalimilan:
```suggestion
    Tables.table(m::AbstractMatrix; [header::Vector{Symbol}])
```

Same below.

src/matrix.jl
```diff
@@ -0,0 +1,71 @@
...
+
+Convert an AbstractMatrix (Matrix, Adjoint, etc.) to a MatrixTable, which satisfies the Tables.jl interface. This allows accesing the matrix via `Tables.rows` and `Tables.columns`. An optional keyword argument `header` can be passed as a Vector{Symbol} to be used as the column names. Note that no copy of the AbstractMatrix is made.
```
nalimilan: Missing backticks around names. I'd also say "wrap" instead of "convert".

Also, how about breaking lines at 92 chars? Text is almost impossible to read currently.

src/matrix.jl
```diff
@@ -0,0 +1,71 @@
...
+function table(m::AbstractMatrix; header::Vector{Symbol}=[Symbol("Column$i") for i = 1:size(m, 2)])
+    length(header) == size(m, 2) || throw(ArgumentError("provided column names `header` length must match number of columns in matrix ($(size(m, 2))"))
+    lookup = Dict(nm=>i for (i, nm) in enumerate(header))
+    return MatrixTable(header, lookup, m)
+end
+
+"""
+Tables.matrix(table)
+
+Convert any table source input to a Matrix. If the table column types are not homogenous, they will be promoted to the most common type in the materialized Matrix. Note that column names are ignored in the conversion.
```
nalimilan: "most common" -> "a common" or "the most appropriate"

Also same remarks as above.

src/matrix.jl
```diff
@@ -0,0 +1,76 @@
...
+
+# column interface
+columnaccess(::Type{<:MatrixTable}) = true
+columns(m::MatrixTable) = m
+Base.getproperty(m::MatrixTable, ::Type{T}, col::Int, nm::Symbol) where {T} = getfield(m, :matrix)[:, col]
+Base.getproperty(m::MatrixTable, nm::Symbol) = getfield(m, :matrix)[:, getfield(m, :lookup)[nm]]
+Base.propertynames(m::MatrixTable) = names(m)
+
+"""
+Tables.table(m::AbstractMatrix; [header::Vector{Symbol}])
```
nalimilan, line 46: Still missing indent.

PR #63, opened 2019-02-03
Use StringVector(undef, dim) constructor

nalimilan:
The other one has been removed from WeakRefStrings 0.5.4 (https://github.com/JuliaData/WeakRefStrings.jl/pull/52).

Unfortunately I don't know if we can mark somewhere that we now require version 0.5.4, since this is an optional dependency.


quinnj (approved):
Thanks!

PR #66, opened 2019-02-08
Finixh #58 by adding the vardim keyword argument to Tables.matrix, al…

quinnj:
…lowing the user to specify whether input columns should be materialized as matrix columns or rows

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -58,20 +58,28 @@ function table(m::AbstractMatrix; header::Vector{Symbol}=[Symbol("Column$i") for
 end
 
 """
-Tables.matrix(table)
+Tables.matrix(table; vardim::Int=2)
 
 Materialize any table source input as a `Matrix`. If the table column types are not homogenous,
 they will be promoted to a common type in the materialized `Matrix`. Note that column names are
 ignored in the conversion.
 """
-function matrix(table)
+function matrix(table; vardim::Int=2)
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks, looks good! My only hesitation is whether we should call the keyword argument `vardim`. The only place where that term is used currently AFAIK is in StatsBase's weighted `cov`, but it's going to be changed to `dims` for consistency with `Statistics.cov`. That said, `vardim` sounds clearer (`dims` would look weird here) and it could be used for https://github.com/JuliaStats/Clustering.jl/issues/79.

andreasnoack:
I think `dims` and `vardim` are equally confusing but since the reductions in Base use `dims`, we have been switching from `vardim` to `dims` in `StatsBase` (although the switch is not yet complete) so I think it would better to use `dims` because of the precedence.

quinnj:
Updated from `vardim` to `dims`.

nalimilan:
That naming issue is quite annoying. After thinking more about it, I feel like using `dims=1` to mean "transform columns to rows" is quite arbitrary. We could as well say `dims=2` means "transform rows to columns", or any other intepretation. There's actually no clear relationship with reductions which use the `dims` argument in Base.

Maybe `transpose::Bool=false` would be clearer? `CSV.file` already supports that argument.

quinnj:
So what should we do here? I don't have strong opinions.

nalimilan:
I think I prefer `transpose`.

tlienart:
I think `transpose` makes sense here. at the end of the day if this is well documented, it should be fine.

Issue #70, opened 2019-02-22
conflicting import of IteratorInterfaceExtensions 


hhaensel:
Compiling `Makie.jl` with `PackageCompiler` leads to a strange error message of `Tables.jl`, as reported
[here](https://discourse.julialang.org/t/ann-packagecompiler-with-incremental-system-images)

Comments by @davidanthoff lead me to removing `using IteratorInterfaceExtensions` in [line 5 of `Tables.jl`](https://github.com/JuliaData/Tables.jl/blob/master/src/Tables.jl#L5) as there is also a [`@require` statement in line 29](https://github.com/JuliaData/Tables.jl/blob/master/src/Tables.jl#L29)

After this modification `Tables.jl` is pre-compiling without error and PackageCompiler does not display any error.

Alternatively, if `IteratorInterfaceExtensions` is required elsewhere, the code in the require statement could be moved to the main part.

nalimilan:
Let's try that then: https://github.com/JuliaData/Tables.jl/pull/71

PR #71, opened 2019-02-22
Remove using IteratorInterfaceExtensions

nalimilan:
It's already done inside `@require` and it breaks package compilation.

Fixes https://github.com/JuliaData/Tables.jl/issues/70.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -2,7 +2,7 @@ module Tables
 
 using Requires, LinearAlgebra
 
-using TableTraits, IteratorInterfaceExtensions
+using TableTraits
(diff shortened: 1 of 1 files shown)


quinnj (approved):

davidanthoff:
That doesn't strike me as the right fix. Wouldn't it make more sense to leave the `using` clause as it was (i.e. let it load `IteratorInterfaceExtensions`), and instead just move everything that is behind the `@require IteratorInterfaceExtensions` statement into the normal module code and remove that whole `@require` clause? I'll open a PR in a minute.

PR #72, opened 2019-02-23
Remove IteratorInterfaceExtensions @require clause

davidanthoff:
I think this is a better fix than https://github.com/JuliaData/Tables.jl/pull/71.

CC @nalimilan

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -4,6 +4,8 @@ using Requires, LinearAlgebra
 
 using TableTraits
 
+import IteratorInterfaceExtensions
+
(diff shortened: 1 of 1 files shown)


nalimilan:
Hmm, so we actually depend on IteratorInterfaceExtensions? Then why was it using `@require`?

davidanthoff:
Yes, and I have no idea.

davidanthoff:
@quinnj Bump, any reason not to merge this?

quinnj (approved):
Thanks!

Issue #88, opened 2019-05-01
Travis tests failing: Unsatistfiable requirements for TableTraits


ablaom:
From [this](https://travis-ci.org/JuliaData/Tables.jl/jobs/525980722) build attempt:

```
ERROR: Unsatisfiable requirements detected for package TableTraits [3783bdb8]:
 TableTraits [3783bdb8] log:
 ├─possible versions are: [0.0.1-0.0.3, 0.1.0, 0.2.0, 0.3.0-0.3.1, 0.4.0-0.4.1, 1.0.0] or uninstalled
 ├─restricted to versions * by Tables [bd369af6], leaving only versions [0.0.1-0.0.3, 0.1.0, 0.2.0, 0.3.0-0.3.1, 0.4.0-0.4.1, 1.0.0]
 │ └─Tables [bd369af6] log:
 │   ├─possible versions are: 0.2.0 or uninstalled
 │   └─Tables [bd369af6] is fixed to version 0.2.0
 ├─restricted to versions 1.0.0 by an explicit requirement, leaving only versions 1.0.0
 └─restricted by compatibility requirements with QueryOperators [2aef5ad7] to versions: [0.0.1-0.0.3, 0.1.0, 0.2.0, 0.3.0-0.3.1, 0.4.0-0.4.1] — no versions left

etc
```

fredrikekre:
The interesting part from the log was left out here, the problem is: https://github.com/JuliaRegistries/General/blob/17357a5c1c15c67fa8ec591c82f3263af6befbef/Q/QueryOperators/Compat.toml#L20
QueryOperators is not compatible with IteratorInterfaceExtensions 1.0

nalimilan:
Looks like it's caused by https://github.com/JuliaLang/Pkg.jl/issues/110. So the only solution is to port QueryOperators to  IteratorInterfaceExtensions 1.0. @davidanthoff Is that going to take a lot of work?

nalimilan:
ActuallyQueryOperators already supports IteratorInterfaceExtensions 1.0: its Project.toml file doesn't have an upper bound, and tests pass. See https://github.com/JuliaRegistries/General/pull/494.

quinnj:
Closing since this isn't a Tables.jl issue.

PR #89, opened 2019-05-02
Use TableTraits.isiterabletable in rows and columns

### LINKED ISSUE #33: DataFrame(load("file.xlsx", "sheetname")) not working
After a recent update in `DataFrames` v0.18, the line `DataFrame(load("file.xlsx", "sheetname"))` no longer works to read an Excel file into a DataFrame. I get the following error:

> ArgumentError: no default `Tables.columns` implementation for type: ExcelFiles.ExcelFile
> 
> Stacktrace:
>  [1] #DataFrame#365(::Bool, ::Type, ::ExcelFiles.ExcelFile) at .julia/packages/Tables/qlc3J/src/fallbacks.jl:156
>  [2] DataFrame(::ExcelFiles.ExcelFile) at .julia/packages/DataFrames/dXFYy/src/other/tables.jl:20
> 

It seems `ExcelFiles` needs to update to use the `Tables` API.

nalimilan:
Otherwise we misdetect some table types which do not define `Base.iterable`, like `ExcelFile`.
Add a new test for this pattern.

Fixes https://github.com/queryverse/ExcelFiles.jl/issues/33, which appeared after https://github.com/JuliaData/Tables.jl/pull/84.

CI fails due to version incompatibilities, not sure why (EDIT: that's https://github.com/JuliaData/Tables.jl/issues/88).

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -58,7 +58,7 @@ function rows(x::T) where {T}
     if columnaccess(T)
         cols = columns(x)
         return RowIterator(cols, rowcount(cols))
-    elseif Base.isiterable(T)
+    elseif TableTraits.isiterabletable(x) === true || Base.isiterable(T)
         return IteratorWrapper(IteratorInterfaceExtensions.getiterator(x))
     end
     throw(ArgumentError("no default `Tables.rows` implementation for type: $T"))
@@ -149,7 +149,7 @@ end
         return buildcolumns(schema(r), r)
     elseif TableTraits.supports_get_columns_copy_using_missing(x)
         return TableTraits.get_columns_copy_using_missing(x)
-    elseif Base.isiterable(T)
+    elseif TableTraits.isiterabletable(x) === true || Base.isiterable(T)
         iw = IteratorWrapper(IteratorInterfaceExtensions.getiterator(x))
         return buildcolumns(schema(iw), iw)
     end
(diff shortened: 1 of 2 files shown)


quinnj:
Thanks!

src/fallbacks.jl
```diff
@@ -58,7 +58,7 @@ function rows(x::T) where {T}
     if columnaccess(T)
         cols = columns(x)
         return RowIterator(cols, rowcount(cols))
-    elseif Base.isiterable(T)
+    elseif TableTraits.isiterabletable(x) === true || Base.isiterable(T)
```
davidanthoff, line 61: I think we could further simplify this: `TableTraits.isiterabletable(x)` will return `true` if `Base.isiterable(T)` is `true`, so we don't need that second test. Also, I think we also want to take this branch if `TableTraits.isiterabletable(x)===missing`, right? So I think the test should probably read `elseif TableTraits.isiterabletable(x) !== false`.

But having said this, this PR is clearly fixing a bug already, so thanks for opening it!

nalimilan: You're clearly the expert here, so feel free to open a PR to improve this.

davidanthoff: Will do once I'm done with my current travel stint. @quinnj could you tag a bug fix release that includes this PR soon? Tables v0.2.0 broke the entire file IO story in Queryverse, and this PR fixes that.

PR #91, opened 2019-05-04
Simplify the fallback logic a bit while ensuring any object that is I…

### LINKED ISSUE #1780: Error in conversion from ExcelFile
I have existing code that used to run without errors. It reads an Excel file using FileIO.jl and converts the result to a DataFrame, which is then processes and an array of structs is returned.

The relevant code is:

```
using FileIO
using DataFrames

function readdata(source)
    data = DataFrame(load(source, "Data"))
    <further processing to create array of structs called emps>
    return emps
 end
```
When called the first time, it errors. When called immediately again, it runs:

```
julia> emps = readdata(".\\data\\Data.xlsx")
ERROR: ArgumentError: unable to construct DataFrame from ExcelFiles.ExcelFile
Stacktrace:
 [1] DataFrame(::ExcelFiles.ExcelFile) at C:\Users\braam\.julia\packages\DataFrames\IKMvt\src\other\tables.jl:23
 [2] readdata(::String) at D:\Braam\Documents\Julia Source\EEMC\src\data_retrieval.jl:33
 [3] top-level scope at none:0

julia> emps = readdata(".\\data\\Data.xlsx")
458-element Array{Employee,1}:
<snip>
```

This seems to be a bug in the current version of DataTables, but I am leaving a lot of room for the error being on my side. Any advise would be appreciated.

quinnj:
…teratorInterfaceExtensions.isiterable is allowed to be iterated. Should fix https://github.com/JuliaData/DataFrames.jl/issues/1780

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -58,7 +58,7 @@ function rows(x::T) where {T}
     if columnaccess(T)
         cols = columns(x)
         return RowIterator(cols, rowcount(cols))
-    elseif TableTraits.isiterabletable(x) === true || Base.isiterable(T)
+    elseif IteratorInterfaceExtensions.isiterable(x)
         return IteratorWrapper(IteratorInterfaceExtensions.getiterator(x))
     end
     throw(ArgumentError("no default `Tables.rows` implementation for type: $T"))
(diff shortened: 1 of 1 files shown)


quinnj:
cc: @nalimilan @davidanthoff I believe this should resolve any TableTraits integration issues.

PR #92, opened 2019-05-08
Implement DataAPI

quinnj:


src/fallbacks.jl
```diff
@@ -59,20 +59,18 @@ function rows(x::T) where {T}
...
 # build columns from rows
-haslength(L) = L isa Union{Base.HasShape, Base.HasLength}
-
 """
     Tables.allocatecolumn(::Type{T}, len) => returns a column type (usually AbstractVector) w/ size to hold `len` elements
 
     Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
 """
-allocatecolumn(T, len) = Vector{T}(undef, len)
+allocatecolumn(T, len) = DataAPI.defaultarray(T, 1)(undef, len)
```
nalimilan, line 74: Should `defaultarray` replace `allocatecolumn` at some point?

src/tofromdatavalues.jl
```diff
@@ -1,24 +1,21 @@
 # utilities for converting Row iterators to and from DataValue-based NamedTuple iterators
 # the DataValue-specific definitions are in the Tables.jl __init__ DataValues @require block
-nondatavaluetype(::Type{T}) where {T} = T
-nondatavaluetype(::Type{Union{}}) = Union{}
-datavaluetype(::Type{T}) where {T} = T
-datavaluetype(::Type{Union{}}) = Union{}
-
-Base.@pure function nondatavaluetype(::Type{NT}) where {NT <: NamedTuple{names}} where {names}
-    TT = Tuple{Any[ nondatavaluetype(fieldtype(NT, i)) for i = 1:fieldcount(NT) ]...}
+DataAPI.nondatavaluetype(::Type{T}) where {T} = T
```
nalimilan: I know that this function is documented to be owned by Tables.jl, but given how simple this fallback definition is, why not put it in DataAPI instead? That would be simpler for everybody.

src/fallbacks.jl
quinnj: they're slightly different; `defaultarray` just returns the array type for a given scalar type; allocatecolumn is for actually creating a column. We can probably just replace uses of `allocatecolumn` with `defaultarray(T)(undef, len)`

src/tofromdatavalues.jl
quinnj: That's fair.

src/fallbacks.jl
nalimilan: Yes, that's what I had in mind.

PR #94, opened 2019-05-15
Add a new CopiedColumns wrapper struct to signal to downstream sinks …

quinnj:
…that the columns source columns have been generated/copied already, to avoid unnecessary copies in sink functions themselves. Addresses https://github.com/JuliaData/DataFrames.jl/issues/1809

cc: @davidanthoff, @nalimilan, @bkamins

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -143,15 +143,28 @@ function _buildcolumns(rowitr, row, st, sch, L, columns, rownbr, len, updated)
     return updated[]
 end
 
+struct CopiedColumns{T}
+    x::T
+end
+
+source(x::CopiedColumns) = getfield(x, :x)
+istable(::Type{<:CopiedColumns}) = true
+columnaccess(::Type{<:CopiedColumns}) = true
+columns(x::CopiedColumns) = x
+schema(x::CopiedColumns) = schema(source(x))
+materializer(x::CopiedColumns) = materializer(source(x))
+Base.propertynames(x::CopiedColumns) = propertynames(source(x))
+Base.getproperty(x::CopiedColumns, nm) = getproperty(source(x), nm)
+
 @inline function columns(x::T) where {T}
     if rowaccess(T)
         r = rows(x)
-        return buildcolumns(schema(r), r)
+        return CopiedColumns(buildcolumns(schema(r), r))
     elseif TableTraits.supports_get_columns_copy_using_missing(x)
-        return TableTraits.get_columns_copy_using_missing(x)
+        return CopiedColumns(TableTraits.get_columns_copy_using_missing(x))
     elseif IteratorInterfaceExtensions.isiterable(x)
         iw = IteratorWrapper(IteratorInterfaceExtensions.getiterator(x))
-        return buildcolumns(schema(iw), iw)
+        return CopiedColumns(buildcolumns(schema(iw), iw))
     end
     throw(ArgumentError("no default `Tables.columns` implementation for type: $T"))
 end
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -154,7 +154,7 @@ columns(x::CopiedColumns) = x
 schema(x::CopiedColumns) = schema(source(x))
 materializer(x::CopiedColumns) = materializer(source(x))
 Base.propertynames(x::CopiedColumns) = propertynames(source(x))
-Base.getproperty(x::CopiedColumns, nm) = getproperty(source(x), nm)
+Base.getproperty(x::CopiedColumns, nm::Symbol) = getproperty(source(x), nm)
 
 @inline function columns(x::T) where {T}
     if rowaccess(T)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -258,7 +258,7 @@ function TableTraits.get_columns_copy_using_missing(x::ColumnSource)
 end
 
 let x=ColumnSource()
-    @test Tables.columns(x) == TableTraits.get_columns_copy_using_missing(x)
+    @test Tables.source(Tables.columns(x)) == Tables.source(Tables.CopiedColumns(TableTraits.get_columns_copy_using_missing(x)))
 end
 
 struct ColumnSource2
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -271,7 +271,7 @@ IteratorInterfaceExtensions.getiterator(::ColumnSource2) =
     Tables.rows((a=[1,2,3], b=[4.,5.,6.], c=["A", "B", "C"]))
 
 let x=ColumnSource2()
-    @test Tables.columns(x) == (a=[1,2,3], b=[4.,5.,6.], c=["A", "B", "C"])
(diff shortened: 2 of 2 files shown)


davidanthoff:
Yes, I think that is probably enough to fix the immediate problem in JuliaData/DataFrames.jl#1809.

I do think that the root problem here is a different one, though: I think a column API should give the consumer control over whether copies are made or not on the source side. This PR helps the client to figure out after the fact whether a copy was made or not, but I think a more robust design is to put the client in control. What if I'm a client who doesn't want a copy to be made? If I use `Tables.columns`, I can now learn after the fact whether a copy was made or not, but that is not ideal, because I might not have called `Tables.columns` in the first place if I had known that copies will be created. This "client is in control" principle is what guided me in the column interface design of TableTraits.jl last year, and I still think that is a good design in this situation.

quinnj:
Well, the scenario here is a bit different than what you're describing: `Tables.CopiedColumns` will only be returned when Tables.jl itself knows copies were already made based on the source type, so currently it's more of a signal that copies **had** to be made to get the columns in the first place. The reason I like this approach is that we don't have to get Tables.jl in the business of making copies or not, but that can be left to clients. Like DataFrames.jl, for example, would like to, by default, "own" the columns inside a DataFrame; w/ this change, they now know after calling `Tables.columns` whether or not the result will be "ownable" or whether they should make additional copies. I feel like this is the minimal change as part of the tables interface, while still allowing clients to do everything they want, making copies or whatever. At least to my knowledge, it's only in-memory tables that really care about this scenario anyway, so not complicating the rest of the API for all the non-in-memory sinks seems like a big win.

With respect to the TableTraits design, I didn't realize it was "fully baked"; I remember first hearing about the idea of `supports_get_columns_copy`, but then didn't really hear much else and, at least as far as I can search on GitHub, it's not widely implemented (only CSVFiles.jl and FeatherFiles.jl as far as I can tell). It also wasn't clear to me why there was only `supports_get_columns_copy_using_missing` and not a corresponding version for DataValues? I also couldn't find any implementations of `TableTraits.get_columns_view`.

quinnj:
Which makes me wonder if a better name here would be `Tables.OwnableColumns`, to make it clear that the columns being returned by `Tables.columns` are safe for the sink to "own".

davidanthoff:
I think this PR makes things work for clients that want to own columns. I don't see how it works well for clients that would want to get a column view, but want to avoid a copy (i.e. rather use row iteration if getting a column would imply a copy). But I think on net this PR is an improvement, because the case of a client who wants to avoid copies being made is also not well handled by the current `master` implementation, so on that front this PR doesn't really change anything.

> With respect to the TableTraits design, I didn't realize it was "fully baked"

TableTraits.jl has essentially been stable and done for almost two years now (with the only small breaking change aligning with the julia 1.0 release). I occasionally add things, always in a non-breaking way. When I add something, I also consider it "done" and won't break anything going forward. The recent version number increase to 1.0.0 really could have happened when julia 1.0 shipped. Not all interfaces are currently used by shipping code, but I consider them stable nevertheless and they are all motivated by various experimental stuff I'm working on.

quinnj:
> I don't see how it works well for clients that would want to get a column view, but want to avoid a copy (i.e. rather use row iteration if getting a column would imply a copy)

I think this is addressed thanks to one of the strengths of the Tables.jl interface: as a client, you don't have to worry whether the source is row-based or column-based; a client just calls `Tables.rows` or `Tables.columns` according to which paradigm better fits the ***sink***. If the client wants rows, they just call `Tables.rows` and they'll always get rows to iterate over, which by its very nature doesn't deal w/ whether copies were made or not. So, as a client, instead of worrying about whether `Tables.columns` will make a copy or not, I should just use `Tables.rows` or `Tables.columns` according to which better fits my sink paradigm. If I do want to use `Tables.columns`, ***and*** I'm worried about having complete ownership of the column vectors, then I should make copies myself; and this PR just lets the sink know that there are a few cases where the columns are already coming through as effectively "copied", so they can avoid the extra copy.

quinnj:
(Oh, the other reason I thought the column interface wasn't "fully baked" is because the documentation still says it's experimental: http://www.david-anthoff.com/jl4ds/stable/tabletraits/#The-columns-copy-interface-[experimental]-1)

nalimilan:
Makes sense, but I wonder whether a type is the best option here: that will force compiling functions for `Columns` and `CopyColumns` even though in most cases the code will be largely similar or even identical. Why not make this a Boolean field of `Columns`?

quinnj:
No, it doesn't force compiling any more than it did before; it's purely a wrapper type `struct CopiedColumns{T}`, so if the result before was unique and forced recompilation, it's still the same now. I'm also not sure what you mean by this `Columns` type?

nalimilan:
What I mean is that functions will be compiled both for `T` and `CopiedColumns{T}`. But I forgot `Tables.columns` returns a named tuple by default, and not a custom type (what I called `Columns`), so there's no way to store that Boolean.

davidanthoff:
Sure, the situation where the client wants column copies is handled fine with this PR. What I pointed out is that a client who wants a column view (and specifically _not_ a column copy) doesn't have that option with the current design.

quinnj:
I'm not sure I understand or at least can imagine what this scenario would be; do you have an example you can share? Also, I still feel like we're maybe not using quite the same ideas; I know this wrapper struct is called `CopiedColumns`, would as I mentioned, it could really just be called `OwnableColumns`, because, at least in the `buildcolumns` case, there ***isn't*** actually any copying going on, we're just materializing vectors from the input source rows.

So to me, the scenario you're describing doesn't really make sense; if you want the data, you want the data, you could have called `Tables.rows` yourself if you wanted to avoid materializing columns, but if you're calling `Tables.columns`, it seems you want materialized vectors anyway. I mean, there's a possibility if the input source supported random access that you could create a `ColumnView` type that just wrapped the input row source and provided an indirect `getindex` that accessed the original row data.

Anyway, I think maybe we're just not speaking apples to apples here.

asinghvi17:
Just an FYI - this breaks things that were relying on the result of `Tables.columns` to be iterable.  This can be fixed by wrapping the output in `Tables.source`.

quinnj:
Please don't rely on the result of `Tables.columns` being iterable; we've tried very hard to document that `Tables.columns` returns a ***property-accessible object of iterables***; it's a core principle of the entire Tables.jl interface. Meaning, things were already broken if you were relying on the result being iterable.

If you need something like a `NamedTuple` of `AbstractVector`s in your application (which happens to iterate over columns), then I'd suggest using `Tables.columntable(x)` directly, which calls `Tables.columns` and then transforms the result to a `NamedTuple` of `AbstractVector`s. 

I'm happy to discuss/consult on the specific use-case if you have more details.

asinghvi17:
Thanks @quinnj!  I was trying to fix the StatsMakie example test failures (the `Group` feature was failing with that error) and saw that change in the tests section of `Tables.jl`.  I've changed it to use `columntable` now, and it passes tests.  

Ref: JuliaPlots/StatsMakie.jl#77 and JuliaPlots/StatsMakie.jl#78

PR #97, opened 2019-05-23
Slightly simplify README MyTable loop

iamed2:
I think this makes it read more like a standard iteration loop (which it is).


quinnj (approved):
Thanks!

PR #98, opened 2019-05-27
support julia 0.7

bjarthur:
this should work once https://github.com/queryverse/IteratorInterfaceExtensions.jl/pull/3 get's merged and tagged, and is needed for DataFrames to continue to support julia 0.7. see https://github.com/JuliaData/DataFrames.jl/pull/1784#issuecomment-493694454.  i know not everyone thinks that that is useful, but i do, and since the effort involved is so trivial, i'm happy to take in on myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ os:
   - linux
   - osx
 julia:
+  - 0.7
   - 1.0
   - 1.1
   - nightly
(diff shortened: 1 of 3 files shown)


nalimilan (approved):
Sounds simple enough, I don't see any reason not to support 0.7 if it already works (for now at least).

quinnj:
It's been over 6 months and no one else has requested this, so closing for now.

PR #105, opened 2019-07-07
Improve performance for known schema column building

quinnj:
cc: @davidanthoff

### DIFF
--- a/src/tofromdatavalues.jl
+++ b/src/tofromdatavalues.jl
@@ -32,7 +32,7 @@ Base.size(rows::IteratorWrapper) = size(rows.x)
 
 @noinline invalidtable(::T, ::S) where {T, S} = throw(ArgumentError("'$T' iterates '$S' values, which don't satisfy the Tables.jl Row-iterator interface"))
 
-function Base.iterate(rows::IteratorWrapper)
+@inline function Base.iterate(rows::IteratorWrapper)
     x = iterate(rows.x)
     x === nothing && return nothing
     row, st = x
@@ -40,7 +40,7 @@ function Base.iterate(rows::IteratorWrapper)
     return IteratorRow(row), st
 end
 
-function Base.iterate(rows::IteratorWrapper, st)
+@inline function Base.iterate(rows::IteratorWrapper, st)
     x = iterate(rows.x, st)
     x === nothing && return nothing
     row, st = x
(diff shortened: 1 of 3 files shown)


src/tofromdatavalues.jl
```diff
@@ -78,21 +78,50 @@ Base.IteratorSize(::Type{DataValueRowIterator{NT, S}}) where {NT, S} = Base.Iter
...
-        end
-        # @show q
-        return q
-    else
-        x = iterate(rows.x, st...)
-        x === nothing && return nothing
-        row, st = x
-        return NT(Tuple(convert(fieldtype(NT, i), getproperty(row, DataValueInterfaces.nondatavaluetype(fieldtype(NT, i)), i, names[i])) for i = 1:fieldcount(NT))), (st,)
-    end
+struct DataValueRow{NT, T}
```
quinnj: @davidanthoff, this change might need discussion. Here, I introduce a lazy `DataValueRow` type that gets iterated instead of a fully-materialized NamedTuple. In the benchmark we've been discussing, it took the time from 560 microseconds down to 290 microseconds by avoiding generating the full NamedTuple. Currently, we make the Enumerable* family happy by declaring the eltype of `DataValueRowIterator` to be the NamedTuple type, and define the appropriate `convert` function if it ever needs to be fully materialized. This essentially allows any DataValue wrapping to only occur on values that are actually accessed. As far as I can tell in some stress testing, all the Enumerable* functions/iterators seem to be happy w/ using this.

davidanthoff: I think that would break a lot of things, as of right now: I just tried to materialize a normal query into some other queryverse sinks, and that fails (e.g. QueryTables.jl), and also things like `@select` break. Haven't tried much more, but the assumption that things are `NamedTuple`s is baked pretty deeply into the wider queryverse right now. Query.jl itself is probably the least problematic, because at least the non-tabular query operators all don't really care what the element type is they operate on, but beyond that small set of operators, I think things would break widely.

I've been mulling whether I should relax the named tuple requirement in general and go with the property accessible story that you use in Tables in general for TableTraits, but I'm still thinking it through and am undecided at this point :) But if we were to go with that in TableTraits, I think we would need to phase it in by first changing all sinks to accept that, and then starting to change the sources.

So I think at least in the very short run we should probably just not go with this, but maybe you could put that code in a PR, so that we can keep experimenting with it?

quinnj: Yeah, I was wondering as such, thanks for looking into it, I'll take this out for now.

quinnj: Here's a branch for experimenting: https://github.com/JuliaData/Tables.jl/pull/106

PR #110, opened 2019-07-12
Make sure we account for DataValue{Any}, which doesn't quite fit in t…

### LINKED ISSUE #109: DataValue{Any} is not converted to missing
This is with v0.2.8:
```julia
julia> DataFrame([(a=DataValue{Any}(), b=DataValue{Int}())])
1×2 DataFrame
│ Row │ a   │ b       │
│     │ Any │ Int64⍰  │
├─────┼─────┼─────────┤
│ 1   │ #NA │ missing │
```
Here column `a` should be of type `Union{Any,Missing}`, not `Any`, and the value in (1,1) should be `missing`, not `NA`.

quinnj:
…he normal convert-to-nondatavaluetype paradigm. Here, we call nondatavaluetype, and if the result is a different type, we know the value we're dealing w/ is a DataValue. We then call a new method, unwrap, if it's a DataValue, which has a special overload for Any. If it's not a DataValue type, no conversion is needed. Fixes #109

cc: @davidanthoff @nalimilan 

I think this is a pretty good solution that is still compiler-friendly (i.e. the non-DataValue case will be optimized away)

### DIFF
--- a/src/tofromdatavalues.jl
+++ b/src/tofromdatavalues.jl
@@ -19,7 +19,7 @@ Tables.rowaccess(::Type{<:IteratorWrapper}) = true
 Tables.rows(x::IteratorWrapper) = x
 
 function Tables.schema(dv::IteratorWrapper{S}) where {S}
-    eT = eltype(S)
+    eT = eltype(dv.x)
     (!(eT <: NamedTuple) || eT === Union{}) && return schema(dv.x)
     return Tables.Schema(nondatavaluenamedtuple(eT))
 end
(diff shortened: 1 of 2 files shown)


davidanthoff:
Yep, this looks good!

src/tofromdatavalues.jl
```diff
@@ -52,17 +52,26 @@ struct IteratorRow{T}
     row::T
 end
 
+unwrap(::Type{T}, x) where {T} = convert(T, x)
+unwrap(::Type{Any}, x) = x.hasvalue ? x.value : missing
```
davidanthoff, line 56: Maybe replace this with `get(x, missing)`? That way you don't rely on the internal structure of `DataValue` (which might change in the future), and I think that should amount to the same code running in the end.

PR #112, opened 2019-07-24
Add Tables.columnindex and Tables.columntype that work on tables them…

quinnj:
…selves

cc: @bkamins


bkamins:
We might have a problem when schema is `nothing`?

quinnj:
I mean, there's not much we can do in that case, so an error seems appropriate.

PR #113, opened 2019-07-26
Add lazy Filter operation

quinnj:
This is pretty simple, so maybe we should have it (it's been requested).


nalimilan:
So our policy is that elementary operations are implemented in a lazy form in this package? Just to make sure we have a consistent rule.

PR #119, opened 2019-09-04
Add rename and rename! to Tables.jl API

bkamins:
Following https://github.com/JuliaData/DataFrames.jl/issues/1514#issuecomment-527935887.

The open question is if we want both `rename` and `rename!` in the common API (`rename` is probably more universally needed, `rename!` is applicable in DataFrames.jl but not in contexts where table does not allow changing column names in-place). I propose to have both but please comment (`rename!` can also live in DataFrames.jl only otherwise)

CC @piever - for syncing with JuliaDB.jl.

### DIFF
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -130,3 +130,6 @@ end
     row, st = state
     return SelectRow{typeof(row), names}(row), st
 end
+
+function rename end
+function rename! end


nalimilan:
How do we decide whether it should be here or in DataAPI? Anyway, it would be nice to have docstrings so that the interface is the same across packages.

quinnj:
In my mind, `rename` is a pretty tables-specific thing, and there are a bunch of packages that already have a dependency on Tables.jl (vs. DataAPI), so it makes sense to me to have it here.

bkamins:
I left out docstrings on purpose. The reason is that e.g. in DataFrames.jl the target API will be:
```
  rename(df::AbstractDataFrame, (from => to)::Pair{Symbol, Symbol}...)
  rename(df::AbstractDataFrame, d::AbstractDict{Symbol,Symbol})
  rename(df::AbstractDataFrame, d::AbstractArray{Pair{Symbol,Symbol}})
  rename(f::Function, df::AbstractDataFrame)
  rename(df::AbstractDataFrame, AbstractVector{Symbol}, makeunique::Bool=false)
```
and I was afraid that it might not be commonly accepted. @nalimilan - do you have an opinion what would be the best course of action here.

Other thing - I understand - given no comment on this issue - that it is OK to have both `rename` and `rename!` defined here.

A complementary note - it seems that JuliaDB.jl has a different API (I have not dug into it in detail but it is different for sure).

piever:
From what I remember JuliaDB used to support more options but now it's just:
```
rename(t, (from => to)::Pair{Union{Int, Symbol}, Symbol}...)
```
or at least that what it looks like from the [docs](https://github.com/JuliaComputing/IndexedTables.jl/blob/master/src/columns.jl#L568).

There is a minor concern of type-piracy in that IndexedTables defines `rename` without restricting the types, and indeed it works on the various table types there as well as on StructArrays. So IndexedTables would need to overload `rename` for its own table types as well as StructArrays, which is not ideal. The same concerns apply to `transfrom` and `select`, in case we later want to unify those.

In my vision of the dependency system, StructArrays has Tables via Requires but I would like to get rid of both eventually and take only a DataAPI dependency. StructArrays unfortunately does not really have the machinery to implement `select`, `transform` and `rename` (even though just `rename` alone is probably not too bad).

nalimilan:
OK, I guess we can just document the minimal signature using pairs. DataFrames can still provide other convenience methods.

> There is a minor concern of type-piracy in that IndexedTables defines `rename` without restricting the types, and indeed it works on the various table types there as well as on StructArrays. So IndexedTables would need to overload `rename` for its own table types as well as StructArrays, which is not ideal. The same concerns apply to `transfrom` and `select`, in case we later want to unify those.

That sounds bad indeed. How about putting a generic definition in Tables.jl instead?

> In my vision of the dependency system, StructArrays has Tables via Requires but I would like to get rid of both eventually and take only a DataAPI dependency. StructArrays unfortunately does not really have the machinery to implement `select`, `transform` and `rename` (even though just `rename` alone is probably not too bad).

So what do you suggest? Defining empty functions in DataAPI?

piever:
> How about putting a generic definition in Tables.jl instead?

I also think the best way forward is to keep a fallback that is efficient for column-wise storage in Tables.jl, so that StructArrays does not need to implement extra logic and can hope to remove Requires dependency (there were already some complaints that StructArrays should be more lightweight). I imagine that testing whether the table has column access, then change the names and call `materialize` should do it. `select` and `transform` may be trickier, but we can worry about that later.

Drvi:
I'd also appreciate `rename` and `rename!` living here (some column-oriented implementation as fallback would make sense to me). Btw, I found this comment on JuliaDB way of doing `rename!` https://discourse.julialang.org/t/juliadb-column-names/24250).

quinnj:
What's the status/consensus here @bkamins, @nalimilan, @piever; should it live here? Should it live in DataAPI? Do we have the right methods worked out/decided w/ generic definitions living here?

bkamins:
We have the API settled in DataFrames.jl for 1.0 release, so @piever - could you please comment how you see it from JuliaDB.jl perspective. Thank you!

piever:
In the end I feel that StructArrays will be fine with depending on Tables (esp. now that it is much more lightweight), so there are no big issues. In my previous comment I was inaccurate in that IndexedTables does indeed only document `rename(t, (from => to)::Pair{Union{Int, Symbol}, Symbol}...)`, but also (by looking at the code) allows `rename(t, p)` with `p` an iterator of pairs. Is this consistent with DataFrames?

I would tend to prefer having a fallback here (potentially with the vararg and the pair iterator options), but would be fine also having only the definitions here (esp. since adding a fallback is technically non-breaking). I suspect it'd be helpful (in the fullness of time) to have a "basic dict-like untyped table" here that can be used for this type of transformations.

bkamins:
We support:
```
  rename(df::AbstractDataFrame, vals::AbstractVector{Symbol}; makeunique::Bool=false)
  rename(df::AbstractDataFrame, vals::AbstractVector{<:AbstractString}; makeunique::Bool=false)
  rename(df::AbstractDataFrame, (from => to)::Pair...)
  rename(df::AbstractDataFrame, d::AbstractDict)
  rename(df::AbstractDataFrame, d::AbstractVector{<:Pair})
  rename(f::Function, df::AbstractDataFrame)
```
so essentially what you have listed is a subset of DataFrames.jl API (except that we accept only `AbstractVector` and not arbitrary iterator).

bkamins:
I am OK to move this issue into another package. My point is that I do not want to have such common names defined in DataFrames.jl namespace, but rather in some umbrella package.

PR #120, opened 2019-11-10
Improve inferrability of IteratorRow

tkf:
This PR makes `hasproperty(IteratorRow((a = 1,)), :a)` inferable.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ DataValueInterfaces = "e2d170a0-9d28-54be-80f0-106bbe20a464"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -13,6 +13,7 @@ DataValueInterfaces = "e2d170a0-9d28-54be-80f0-106bbe20a464"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
quinnj, line 16: Why do we need `Compat` here?

tkf: This is for using `hasproperty` in test.

tkf:
I realized that `hasproperty` is not inferred even for named tuples (except the first entry).  So I guess this fix is not worth the trouble.

PR #128, opened 2020-01-10
Add ~~Interface~~ tests for SparseMatrix

kescobo:
I talked with @quinnj about this at JuliaCon... only just getting to it :-(

I more or less copied the interface for Matrix. A couple of changes were needed for tests, since characters/strings etc aren't supported for sparse matrices.


ablaom:
When we do `Tables.table(A)` and `A` has a lot of columns (is column-sparse) then potentially the tables object carries around a big vector `[:Column1, ..., :Column123498980]` (in the proposed implementation). I wonder if there is a way to avoid that. 

Columns are always called `Column1`, ... right? So could `getproperty(::Tables.SparseMatrixTable, ...)` use this fact in way that makes carrying around the column names unnecessary? Similarly, `schema` could  autogenerate the column names. And so on.

kescobo:
Yeah - I copied all that from the Matrix interface, but I can see it might have different considerations for sparse matrices. That said - most other implementations don't handle sparse columns - are you suggesting that we should trim empty columns? Could maybe add a `skipemtpy=false` or something - should we do this for rows too?

I wonder if it would be worth putting this in as a first pass, and then make the modifications you suggest as an upgrade (mostly because I'm not confident in my ability to do it right given my skill and time constraints).

>Columns are always called Column1, ... right?

In principle, someone can pass a `header` argument to the constructor (took that from Matrix implementation too).

Project.toml
```diff
@@ -4,22 +4,24 @@ authors = ["quinnj <quinn.jacobd@gmail.com>"]
...
+
+[extras]
+DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
+QueryOperators = "2aef5ad7-51ca-5a8f-8e88-e75cf067b44b"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "DataValues", "QueryOperators"]
+test = ["Test", "DataValues", "QueryOperators", "SparseArrays"]
```
quinnj, line 26: This shouldn't be needed since we're adding it as a direct dependency.

src/spmatrix.jl
```diff
@@ -0,0 +1,89 @@
+istable(::Type{<:AbstractSparseMatrix}) = false
+istable(::AbstractSparseMatrix) = false
```
quinnj: This 2nd definition isn't needed. There's a generic fallback that will call the first.

quinnj: We should remove the duplicate definition for matrixes too

quinnj (commented):
This looks mostly good, I just left a few comments. I think the bigger question is whether we actually want to add this overall. I can see it being a bit awkward if people run into issues materializing a monster sparse matrix (as a DataFrame, for example), but then again, this isn't that much code, and it's probably helpful in some workflows. It's probably worth checking if this has any impact on package loading time: that would probably be the biggest worry I personally have. People are pretty sensitive to Tables.jl being as lightweight as possible, so if this has a nontrivial impact, then it might be a deal-breaker.

I'd love to hear @nalimilan, @bkamins, @davidanthoff thoughts on this.

nalimilan:
I don't understand why a special `SparseMatrixTable` wrapper is needed. Most of the code seems to be the same as for `MatrixTable`. If some specific methods can be more efficient by specializing on sparse matrices, then better overload them for `::MatrixTable{<:AbstractSparseMatrix}`.

Regarding `Tables.sparsematrix`, it would sound more Julian to do `Tables.matrix(SparseMatrixCSC, t)` instead. That would be extensible to other array types (e.g. `SparseMatrixCSR` in the future).

kescobo:
>Most of the code seems to be the same as for MatrixTable

That's because it is - I just copy-pasted 😆 . @nalimilan Turns out you're 100% right though - the matrix code is actually generic enough that none of the duplication is needed. I thought that it didn't work because I was trying to add support for a different type, but I was missing the `Tables.table()` conversion 🤦‍♂ .

So let's just consider this PR adding tests for SparseArrays...

Project.toml
kescobo: You're right - though based on discussion below, I've removed the direct dependency and kept this in.

src/spmatrix.jl
kescobo: Removed from matrixes file, now all this code is gone too :-P

kescobo:
>People are pretty sensitive to Tables.jl being as lightweight as possible, so if this has a nontrivial impact, then it might be a deal-breaker.

So presumably this is no longer a concern, right? There's no way adding something as a test dependency would increase load times, is there?

quinnj:
Well, we would still want to consider adding a sink function for SparseMatrix, right? Like, turn any table input into a sparse matrix? Also, the tests here don't seem to do anything SparseMatrix-specific, right?

kescobo:
Sure - I didn't see where in the Matrix code there was a sink function - is that somewhere else? Happy to add that. 

And no, nothing specific in the SparseMatrix tests, other than providing inputs that can be put into sparse matrices (Char and String don't work). I also put in some zeros in the inputs, just to make sure they're handled properly. But other than changing the inputs and outputs, all the tests are identical to those for Matrix.

quinnj:
I added some explicit sparse tests in https://github.com/JuliaData/Tables.jl/pull/145

PR #131, opened 2020-02-01
Tables API enhancement

quinnj:
Ok, here's a proposal for an official 1.0 API interface for Tables.jl. This PR is actually non-breaking in the sense that everything currently relying on the Tables.jl interface and it's current API definition will continue working with this code (see the unchanged, yet passing tests on this PR), but it enhances the API definitions to require a bit more in order to provide additional functionality, conveniences, and consistency. The enhanced APIs mean sources will need to enhance their implementations to be compatible with sinks who start using the new enhanced APIs.

To try and boil the changes down as succinctly as possible, we have:
* For the "Row" interface, from requiring `propertynames(x)` and `getproperty(x, nm)` to requiring `Tables.columnnames(x)` and `Tables.getcolumn(x, col::Union{Int, Symbol})`
* Also requiring the same `columnnames` and `getcolumn` definitions on the "Columns" interface
* When you call `Tables.getcolumn(x, col)` on a "Columns" object, it must return a column as an indexable collection with known length, instead of just as an iterator

The motivation for these changes include:
  * Allowing objects where it's difficult/piracy to overload `propertynames`/`getproperty` to participate safely in the Tables interface
  * More consistency: we will now have the same consistent methods when working with a "Columns" object or "Row" object
  * More functionality: you can now access columns/column values by name or index consistently

The biggest disadvantage, IMO, of these changes is requiring less convenient methods for the interfaces; i.e. before, you could reliably count on doing `row.col1`, but that's not strictly required anymore (though could be strongly encouraged). The counterargument to that is the api enhancements make working _generically_ (like with generic packages) with rows/columns more convenient via consistency/functionality, and for more casual users interested in a specific table/row type, they'll probably be more aware of that types official api (like DataFrames, or that `CSV.Row` documents that it supports property-access).

Along those lines, this PR also defines the `AbstractColumns` and `AbstractRow` abstract types, which allow for custom types to inherit useful behavior in the form of default definitions using the interface methods; i.e. they automatically get indexing, property access, abstractdict interface, iteration, etc. just by relying on the `getcolumn` and `columnnames` interface methods. Hopefully that encourages types to still be functionally robust (and for the most part, more consistent) in the case that users want to use the Tables api directly.

As a final note, this PR hasn't completely updated all the docs, and hasn't implemented tests to cover the new functionality, but it's far enough along that I wanted to get feedback and people's thoughts.


src/Tables.jl
```diff
@@ -8,29 +8,128 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
     Base.getproperty(t::Tuple, i::Int) = t[i]
 end
 
-"Abstract row type with a simple required interface: row values are accessible via `getproperty(row, field)`; for example, a NamedTuple like `nt = (a=1, b=2, c=3)` can access its value for `a` like `nt.a` which turns into a call to the function `getproperty(nt, :a)`"
-abstract type Row end
+"""
+    Tables.AbstractColumns
```
oxinabox, line 37: Can we highlight explict that not all column types inherit from AbstractColumn ?
And that it thus should not be depended on for dispatch. 
That it is just for convieinve of source authors (not sink authors)
(And same for Row)

src/Tables.jl
```diff
@@ -8,29 +8,128 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
+Abstract type provided to allow custom table types to inherit useful and required behavior.
+
+Interface definition:
+| Required Methods | Default Definition | Brief Description |
+| ---------------- | ------------------ | ----------------- |
+| `Tables.getcolumn(table, i::Int)` | getfield(table, i) | Retrieve a column by index |
+| `Tables.getcolumn(table, nm::Symbol)` | getproperty(table, nm) | Retrieve a column by name |
+| `Tables.columnnames(table)` | propertynames(table) | Return column names for a table as an indexable collection |
+| Optional methods | | |
+| `Tables.getcolumn(table, ::Type{T}, i::Int, nm::Symbol)` | Tables.getcolumn(table, nm) | Given a column eltype `T`, index `i`, and column name `nm`, retrieve the column. Provides a type-stable or even constant-prop-able mechanism for efficiency.
```
oxinabox, line 51: Can we include in this PR a test suite method that takes a instance of a subtype and checks it has a correct implementation?

src/operations.jl
```diff
@@ -1,18 +1,23 @@
...
-Base.propertynames(row::TransformsRow) = propertynames(getfield(row, 1))
+getrow(r::TransformsRow) = getfield(r, :row)
+getfuncs(r::TransformsRow) = getfield(r, :funcs)
 
-struct Transforms{C, T, F}
+getcolumn(row::TransformsRow, nm::Symbol) = (getfunc(row, getfuncs(row), nm))(getcolumn(getrow(row), nm))
+getcolumn(row::TransformsRow, i::Int) = (getfunc(row, getfuncs(row), i))(getcolumn(getrow(row), i))
+columnnames(row::TransformsRow) = columnnames(getrow(row))
+
+struct Transforms{C, T, F} <: AbstractColumns
```
oxinabox: Before 1.0 should we take the chance to rename this ? (Or to rename the other?)

src/operations.jl
```diff
@@ -22,10 +27,15 @@ function transform(src::T, funcs::F) where {T, F}
...
-getfunc(row, d::Dict{Int, <:Base.Callable}, nm) = get(d, findfirst(isequal(nm), propertynames(row)), identity)
+getfunc(row, nt::NamedTuple, nm::Symbol) = get(nt, nm, identity)
+getfunc(row, d::Dict{String, <:Base.Callable}, nm::Symbol) = get(d, String(nm), identity)
+getfunc(row, d::Dict{Symbol, <:Base.Callable}, nm::Symbol) = get(d, nm, identity)
+getfunc(row, d::Dict{Int, <:Base.Callable}, nm::Symbol) = get(d, findfirst(isequal(nm), columnnames(row)), identity)
+
+getfunc(row, nt::NamedTuple, i::Int) = i > fieldcount(typeof(nt)) ? identity : getfield(nt, i)
+getfunc(row, d::Dict{String, <:Base.Callable}, i::Int) = get(d, String(columnnames(row)[i]), identity)
+getfunc(row, d::Dict{Symbol, <:Base.Callable}, i::Int) = get(d, columnnames(row)[i], identity)
+getfunc(row, d::Dict{Int, <:Base.Callable}, i::Int) = get(d, i, identity)
```
oxinabox, line 51: Can we drop the dispatch on `Base.Callable` ?

bkamins:
If we make these changes could also adding a "public" method taking a table and returning an iterator of `NamedTuples` representing rows be considered as a part of public Tables.jl API for 1.0 release? In this way you could efficiently (and in type stable way) pretend that your table is a `Tables.rowtable` for iteration purposes.

Currently `Tables.rows` does not give this guarantee (and I think it should not), but can return rows of different type depending on the type of source table, so something like a lazy variant of `Tables.rowtable` would be useful (so that no matter what is the source table you can rely on the fact that you iterate `NamedTuple`s - this is relevant when designing API using Tables.jl).

src/Tables.jl
quinnj: Great call out.

src/operations.jl
quinnj: Or remove it entirely. I know @bkamins and @nalimilan have been hammering on the select/transform API in DataFrames, so if there's a way we could generalize their excellent design work into a generic Tables.jl API, I think it'd be a great starting point for a TablesOperations.jl package or something.

src/operations.jl
quinnj: Does it hurt at all? Or overly restrict?

quinnj:
@bkamins, can you clarify the use-case a little for the public namedtuple iterator? Note that the machinery exists to do type-stable iteration over generic rows already: using `Tables.eachcolumn` and the `getproperty(row, T, i, nm)` (now `getcolumn(row, T, i, nm)`) method, which also gracefully handles extremely large datasets by not over-compiling (it switches to a run-length encoding of column types, and if that's still too much, it switches to a single for-loop).

I just worry about publicly blessing NamedTuples too much in relation to data tasks because of the inevitable production issues people will run into with extremely large datasets, especially when we already have machinery to pretty much achieve the same benefits without.

oschulz:
> you could reliably count on doing row.col1, but that's not strictly required anymore (though could be strongly encouraged)

Yes, please (strongly encouragement, at least), this makes code so much more readable (and writeable) ...

bkamins:
> can you clarify the use-case a little for the public namedtuple iterator?

I agree that for very wide tables one should not use a `NamedTuple` iterator, but there are common cases when you want to transform several (e.g. four as in an example below) columns into a single column using each-row operation, e.g. something like:
```
r -> r.a in r.b ? r.c : r.d
```
and then you apply `f` to all rows of some table. Now what is crucial:
* to allow using `getproperty` in the code
* to allow using `r[1]` (i.e. `getindex`) in the code
* to be able to tell the user what is the contract for `r` above

Currently in DataFrames.jl we say that the contract for `r` is that it is a `NamedTuple` and produce it using `Tables.namedtupleiterator`. We could stop doing this, but the we need to set a different contract which `NamedTuple` meets and for sure I would like not to drop allowing the usage of `getproperty` and `getindex`.

src/Tables.jl
quinnj: The difficulty here in doing a static check of interface implementation is we have generic fallback methods for all the interface methods (so things like NamedTuples and Generators of NamedTuples just work). But I think there's something we could provide to run in your test suite to put your type through the works.

src/operations.jl
oxinabox: Good idea.
Deleting things before 1.0 is a good way to free up real estate for redevelopment.

src/Tables.jl
oxinabox: Yeah. We need to send something though it.
We can't just use `hasmethod`

src/operations.jl
oxinabox: It's overly restrictive.
Doesn't work with functors.

Maybe I want to pass something though a trained NN. `Flux.Chain` is callable, but not a subtype of `Base.Callaeble`

quinnj:
@bkamins, ok, I understand that use-case, but I also wonder why `Tables.rowtable` isn't sufficient? For a limited, known use-case, it seems like using a `Vector` of NamedTuples would be just as useful as a namedtuple iterator, without needing to add to the public API surface area of Tables.jl.

I should clarify that I'm not against including `namedtupleiterator` as an official utility function; just doing some regular push-back to make sure existing solutions/work-arounds aren't adequate.

bkamins:
> it seems like using a `Vector` of `NamedTuple`s would be just as useful as a namedtuple iterator

The problem is that materializing a `Vector` of `NamedTuple`s is potentially expensive (as opposed - as I assume - to using just an iterator that does not allocate).

Just a quick code (not useful, but similar things happen in practice):

```
for i in 1:10^6
    df2 = select(df, :, [:a, :b] => Row(x -> x.a + x.b + i) => :c, copycols=false)
    # and do something with df2
end
```

Now `Tables.rowtable` for a `DataFrame` that has 10^6 rows and 3 columns (a normal use case) takes:
```
julia> @benchmark Tables.rowtable(df)
BenchmarkTools.Trial:
  memory estimate:  221.23 MiB
  allocs estimate:  10998485
  --------------
  minimum time:     417.965 ms (1.66% GC)
  median time:      445.184 ms (2.71% GC)
  mean time:        466.125 ms (6.89% GC)
  maximum time:     601.872 ms (17.16% GC)
  --------------
  samples:          11
  evals/sample:     1
```
which is an overkill (now I see that `Tables.rowtable(Tables.columntable(df))` is much faster - I will make a PR to DataFrames.jl fixing this and CC you).

But even given that it is faster (and allocating less to do):
```
x = Tables.rows(Tables.columntable(df))
Tables.namedtupleiterator(eltype(x), x)
```
than
```
Tables.rowtable(Tables.columntable(df))
```
(and iterating both seems to have the same performance later; actually - surprisingly - iterating `Tables.namedtupleiterator` seems 10% faster in my benchmarks)

that is why I would prefer to have `Tables.namedtupleiterator` exposed with simplified API (now it is hard to use except for library codes, as you have to pass two arguments to it).

docs/make.jl
```diff
@@ -0,0 +1,17 @@
+using Documenter, Tables
+
+makedocs(;
+    modules=[Tables],
+    format=Documenter.HTML(),
+    pages=[
+        "Home" => "index.md",
+    ],
+    repo="https://github.com/JuliaData/Tables.jl/blob/{commit}{path}#L{line}",
```
nalimilan, line 9: Is this needed? It's not set up in other packages I know, and the link works.

docs/make.jl
```diff
@@ -0,0 +1,17 @@
+using Documenter, Tables
+
+makedocs(;
+    modules=[Tables],
+    format=Documenter.HTML(),
```
nalimilan, line 5: This isn't necessary AFAIK. Same for `assets`.

docs/Project.toml
```diff
@@ -0,0 +1,2 @@
+[deps]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
nalimilan, line 2: I think it's recommended to specify a particular version to avoid breakage. Then the Manifest isn't needed.

src/utils.jl
```diff
@@ -22,23 +22,19 @@ Base.@pure function runlength(::Type{T}) where {T <: Tuple}
...
+The first definition takes a function `f`, table schema `sch`, a `row` type (that satisfies the `Row` interface), and any other `args...`;
+it generates calls to get the value for each column in the row (`Tables.getcolumn(row, nm)`) and then calls `f(val, col, name, args...)`, where `f` is the
+user-provided function, `val` is a row's column value, `col` is the column index as an `Int`, and `name` is the row's column name as a `Symbol`.
 
-    While the first definition applies to a `Row` object, the last definition simply returns a property-iterator over a `Columns` object.
-    For example, one could "collect" every column of a `Columns` object by doing:
-    ```julia
-    vectors = [collect(col) for col in Tables.eachcolumn(Tables.columns(x))]
-    ```
+While the first definition applies to an `Row` object, the last definition simply returns an AbstractColumn iterator for a `Columns` object.
```
nalimilan:
```suggestion
While the first definition applies to a `Row` object, the last definition simply returns an `AbstractColumn` iterator for a `Columns` object.
```

src/tofromdatavalues.jl
```diff
@@ -48,38 +56,40 @@ end
...
 
 # DataValueRowIterator wraps a Row iterator and will wrap `Union{T, Missing}` typed fields in DataValues
 struct DataValueRowIterator{NT, S}
     x::S
 end
 
+"""
+    Tables.datavaluerows(x) => NamedTuple iterator
+
+Takes any table input `x` and returns a NamedTuple iterator
```
nalimilan:
```suggestion
Takes any table input `x` and returns a `NamedTuple` iterator
```

src/tofromdatavalues.jl
```diff
@@ -48,38 +56,40 @@ end
...
+that will replace missing values with DataValue-wrapped values;
```
nalimilan:
```suggestion
that will replace missing values with `DataValue`-wrapped values;
```

src/tofromdatavalues.jl
```diff
@@ -13,6 +13,14 @@ struct IteratorWrapper{S}
     x::S
 end
 
+"""
+    Tables.nondatavaluerows(x)
+
+Takes any Queryverse-compatible NamedTuple iterator source and
```
nalimilan:
```suggestion
Takes any Queryverse-compatible `NamedTuple` iterator source and 
```

src/tofromdatavalues.jl
```diff
@@ -13,6 +13,14 @@ struct IteratorWrapper{S}
     x::S
 end
 
+"""
+    Tables.nondatavaluerows(x)
+
+Takes any Queryverse-compatible NamedTuple iterator source and 
+converts to a Tables.jl-compatible Row iterator. Will automatically
```
nalimilan:
```suggestion
converts to a Tables.jl-compatible `Row` iterator. Will automatically
```

src/tofromdatavalues.jl
```diff
@@ -13,6 +13,14 @@ struct IteratorWrapper{S}
...
+unwrap any `DataValue`s, replacing `NA` with `missing`.
+Useful for translating Query.jl results back to non-DataValue-based tables.
```
nalimilan:
```suggestion
Useful for translating Query.jl results back to non-`DataValue`-based tables.
```

src/operations.jl
```diff
@@ -1,18 +1,23 @@
...
+struct Transforms{C, T, F} <: AbstractColumns
     source::T
     funcs::F # NamedTuple of columnname=>transform function
 end
 
-Base.propertynames(t::Transforms{true}) = propertynames(getfield(t, 1))
-Base.getproperty(t::Transforms{true}, nm::Symbol) = Base.map(getfunc(t, getfield(t, 2), nm), getproperty(getfield(t, 1), nm))
+columnnames(t::Transforms{true}) = columnnames(getfield(t, 1))
+getcolumn(t::Transforms{true}, nm::Symbol) = Base.map(getfunc(t, getfield(t, 2), nm), getcolumn(getfield(t, 1), nm))
+getcolumn(t::Transforms{true}, i::Int) = Base.map(getfunc(t, getfield(t, 2), nm), getcolumn(getfield(t, 1), i))
```
nalimilan:
```suggestion
getcolumn(t::Transforms{true}, i::Int) = Base.map(getfunc(t, getfield(t, 2), i), getcolumn(getfield(t, 1), i))
```
(Probably needs a test.)

src/operations.jl
nalimilan: Maybe just mention in the docs that this isn't part of the API, and ensure it's unexported? Then we can decide what to do in the long run.

src/namedtuples.jl
```diff
@@ -68,27 +83,43 @@ istable(::Type{<:ColumnTable}) = true
...
+materializer(x::ColumnTable) = columntable
+
 getarray(x::AbstractArray) = x
 getarray(x) = collect(x)
 
+"""
+    Tables.columntable(x) => NamedTuple of Vectors
+    Tables.columntable(ct, x) => ct
+
+Takes any input table source `x` and returns a NamedTuple of Vectors,
```
nalimilan:
```suggestion
Takes any input table source `x` and returns a `NamedTuple` of `Vector`s,
```

src/matrix.jl
```diff
@@ -14,37 +13,37 @@ istable(::Type{<:MatrixTable}) = true
...
-Base.getproperty(m::MatrixTable, nm::Symbol) = getfield(m, :matrix)[:, getfield(m, :lookup)[nm]]
-Base.propertynames(m::MatrixTable) = names(m)
+getcolumn(m::MatrixTable, ::Type{T}, col::Int, nm::Symbol) where {T} = getfield(m, :matrix)[:, col]
+getcolumn(m::MatrixTable, nm::Symbol) = getfield(m, :matrix)[:, getfield(m, :lookup)[nm]]
+getcolumn(m::MatrixTable, i::Int) = getfield(m, :matrix)[:, i]
+columnnames(m::MatrixTable) = names(m)
 
 """
-Tables.table(m::AbstractMatrix; [header::Vector{Symbol}])
+  Tables.table(m::AbstractMatrix; [header::Vector{Symbol}])
```
nalimilan:
```suggestion
    Tables.table(m::AbstractMatrix; [header::Vector{Symbol}])
```

src/matrix.jl
```diff
@@ -58,7 +57,7 @@ function table(m::AbstractMatrix; header::Vector{Symbol}=[Symbol("Column$i") for
 end
 
 """
-Tables.matrix(table; transpose::Bool=false)
+  Tables.matrix(table; transpose::Bool=false)
```
nalimilan:
```suggestion
    Tables.matrix(table; transpose::Bool=false)
```

src/fallbacks.jl
```diff
@@ -153,14 +180,24 @@ end
...
-struct CopiedColumns{T}
+"""
+    Tables.CopiedColumns
+
+For some sinks, there's a concern about whether they can safely "own" columns from the input.
+To be safe, they should always copy input columns, to avoid unintended mutation.
+When we've called buildcolumns, however, Tables.jl essentially built/owns the columns,
+and it's happy to pass ownership to the sink. Thus, any built columns will be wrapped
+in a CopiedColumns struct to signal to the sink that essentially "a copy has already been made"
+and they're safe to assume ownership
```
nalimilan:
```suggestion
and they're safe to assume ownership.
```

src/fallbacks.jl
```diff
@@ -153,14 +180,24 @@ end
...
     return NamedTuple{Base.map(Symbol, names)}(_buildcolumns(rowitr, row, st, sch, columns, Ref{Any}(columns))[])
 end
 
-struct CopiedColumns{T}
+"""
+    Tables.CopiedColumns
+
+For some sinks, there's a concern about whether they can safely "own" columns from the input.
+To be safe, they should always copy input columns, to avoid unintended mutation.
+When we've called buildcolumns, however, Tables.jl essentially built/owns the columns,
```
nalimilan:
```suggestion
When we've called `buildcolumns` however, Tables.jl essentially built/owns the columns,
```

src/fallbacks.jl
```diff
@@ -65,21 +75,29 @@ function Base.iterate(rows::RowIterator, st=1)
...
 
-# build columns from rows
+# for Rows objects, we define a "collect"-like routine to build up columns from iterated rows
+
 """
-    Tables.allocatecolumn(::Type{T}, len) => returns a column type (usually AbstractVector) w/ size to hold `len` elements
+    Tables.allocatecolumn(::Type{T}, len) => returns a column type (usually AbstractVector) with size to hold `len` elements
 
     Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
+    Alternatively, and more generally, custom scalars can overload `DataAPI.defaultarray` to signal the default array type
```
nalimilan:
```suggestion
    Alternatively, and more generally, custom scalars can overload `DataAPI.defaultarray` to signal the default array type.
```

src/Tables.jl
```diff
@@ -8,98 +8,277 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
+  * Iteration interface defined; i.e. `for col in table` will iterate each column in the table
+  * A default `show` method
+This allows a custom table type to behave as close as possible to a builtin `NamedTuple` of vectors object.
+"""
+abstract type AbstractColumns end
 
-First:
-* `Tables.istable(::Type{<:MyTable}) = true`: this provides an explicit affirmation that your type implements the Tables interface
+"""
+    Row
```
nalimilan: I don't understand the reason for having this type. It's confusing to have both `Row` and `AbstractRow`, the two being... abstract. If the point is just to document the `Row` *trait*, then better just mention this in the manual (and in the `AbstractRow` docstring (which already says inheriting isn't needed, but documents the interface anyway).

Also:
```suggestion
    Tables.Row
```

src/Tables.jl
```diff
@@ -8,98 +8,277 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
-end
-```
+# default definitions
+"""
+    Tables.istable(x) => Bool
 
-Obviously every table type is different, but via a combination of `Tables.rows` and `Tables.columns` each table type should be able to construct an instance of itself.
+Check if an object has specifically defined that it is a table. Note that 
+not all valid tables will return true, since it's possible to satisfy the
+Tables.jl interface at "run-time", e.g. a Generator of NamedTuples iterates
```
```suggestion
Tables.jl interface at "run-time", e.g. a `Generator` of `NamedTuples` iterates
```

src/Tables.jl
```diff
@@ -8,98 +8,277 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
+NamedTuples, which satisfies the Row interface, but there's no static way
```
```suggestion
`NamedTuples`, which satisfies the `Row` interface, but there's no static way
```

src/Tables.jl
```diff
@@ -8,98 +8,277 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
-# default definitions
 istable(x::T) where {T} = istable(T) || TableTraits.isiterabletable(x) === true
 istable(::Type{T}) where {T} = false
+
+"""
+    Tables.rowaccess(x) => Bool
+
+Check whether an object has specifically defined that it implements the `Tables.rows`
+function. Note that `Tables.rows` will work on any object that iterates Row-compatible
+objects, even if they don't define `rowaccess`, e.g. a Generator of NamedTuples. Also
```
```suggestion
objects, even if they don't define `rowaccess`, e.g. a `Generator` of `NamedTuples`. Also
```

src/Tables.jl
```diff
@@ -8,98 +8,277 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
-materializer(x) = columntable
+
+"""
+    Tables.materializer(x) => Callable
+
+For a table input, return the "sink" function or "materializing" function that can take a
+Tables.jl-compatible table input and make an instance of the table type. This enables "transform"
+workflows that take table inputs, apply transformations, potentially converting the table to
+a different form, and end with producing a table of the same type as the original input. The
+default materializer is `Tables.columntable`, which converts any table input into a NamedTuple
```
```suggestion
default materializer is `Tables.columntable`, which converts any table input into a `NamedTuple`
```

src/Tables.jl
```diff
@@ -8,98 +8,277 @@ if !hasmethod(getproperty, Tuple{Tuple, Int})
...
+of Vectors.
```
```suggestion
of `Vector`s.
```

docs/src/index.md
```diff
@@ -0,0 +1,202 @@
...
+```@docs
+Tables.rows
+Tables.columns
+```
+
+Given these two powerful data access methods, let's walk through real, albeit somewhat simplified versions of how packages actually use these methods.
+
+### Tables.rows usage
+
+First up, let's take a look at the [SQLite.jl]() package and how it uses the Tables.jl interface to allow loading of generic table-like data into a sqlite relational table. Here's the code:
```
nalimilan: Missing link.

nalimilan (commented):
Some smallish comments and typos.

docs/make.jl
quinnj: This is all generated from PkgTemplates.jl

docs/Project.toml
quinnj: 👍

src/Tables.jl
quinnj: I figured the more docs the better, but yeah, maybe we only need to mention this in the docs.

docs/src/index.md
quinnj: fixed

docs/make.jl
nalimilan: Maybe we got a bit overzealous? Example.jl is much simpler: https://github.com/JuliaLang/Example.jl/blob/master/docs/make.jl

quinnj:
@bkamins, I cleaned up `Tables.namedtupleiterator` internally, added docs, and put it in the official list of "utility functions".

docs/Project.toml
oxinabox: Manifest is useful becaue then you can include in it
```
[[Tables]]
path = ".."
```

which makes sure the docstrings are right for the dev version docs.

src/Tables.jl
tkf: Related to this, I have a thought on interface versioning. See: https://github.com/JuliaData/Tables.jl/issues/133

bkamins:
Thank you. I will add an efficient method for this it in DataFrames.jl when you release this change.

src/namedtuples.jl
```diff
@@ -1,54 +1,112 @@
...
+Take any input table source, and produce a `Vector` of `NamedTuple`s,
+also known as a "row table". A "row table" is a kind of default
+table type of sorts, since it satisfies the Tables.jl row interface
+naturally.
+
+The 2nd definition takes
+an existing row table and appends the input table source `x`
+to the existing row table.
+"""
+function rowtable end
```
tkf, line 80: I suggest to add `isrowtable` trait function #134.  It would introduce "row table" to mean something different.  I'm not sure if you want to include #134 in 1.0.  But can you at least remove `Tables.rowtable` from the public API?  Of course, rejecting #134 and keeping `Tables.rowtable` is a reasonable choice, too.

src/namedtuples.jl
```diff
@@ -1,54 +1,112 @@
...
+    return NamedTuple{names}(Tuple(getcolumn(row, nm) for nm in names)), st
 end
 
-namedtupleiterator(::Type{T}, rows::S) where {T <: NamedTuple, S} = rows
-namedtupleiterator(::Type{T}, rows::S) where {T, S} = NamedTupleIterator{typeof(schema(rows)), S}(rows)
-
 # sink function
+"""
+    Tables.rowtable(x) => Vector{NamedTuple}
+    Tables.rowtable(rt, x) => rt
```
tkf: I don't know enough context to understand why `rowtable(rt, x)` and `columntable(ct, x)` exist. But wouldn't it be better to introduce `tablecat(table, tables...)` or maybe even `Tables.cat(table, tables...)`?

tkf:
I filed a few feature requests that may be easier to resolve before going 1.0:

* Decoupling interface versioning and Tables.jl API versioning? #133
* Feature request: trait `isrowtable` #134
* API clarification request: space/time complexity of `rows(table)` and `columns(table)` #135

quinnj:
@bkamins , just fyi, I just pushed a commit that improves `Tables.namedtupleiterator` performance by 3x (at least in one benchmark I've been running). Just posting in case you have benchmarks you could check as well.

src/namedtuples.jl
quinnj: Yeah, we should probably remove the 2 appending methods; I agree they don't really make sense.

bkamins:
> Just posting in case you have benchmarks you could check as well.

I do not see a significant difference in my benchmarks. What is exactly your test scenario? I was generating `Tables.namedtupleiterator` from a `NamedTuple` of vectors and checked the speed of iterating it (maybe you meant that it is faster to create this iterator rather than iterate it?)

quinnj:
Current master:
```julia
julia> using BenchmarkTools, DataFrames, Tables
[ Info: Precompiling DataFrames [a93c6f00-e57d-5684-b7b6-d8193f3e46c0]

julia> df = DataFrame(a=1:10000, b=1.0:10000.0, c=["hey" for i = 1:10000]);

julia> @btime rowtable(df);
  3.102 ms (117974 allocations: 2.49 MiB)
```
This PR:
```julia
julia> using BenchmarkTools, DataFrames, Tables
[ Info: Precompiling DataFrames [a93c6f00-e57d-5684-b7b6-d8193f3e46c0]

julia> df = DataFrame(a=1:10000, b=1.0:10000.0, c=["hey" for i = 1:10000]);

julia> @btime rowtable(df);
  1.632 ms (97975 allocations: 1.88 MiB)
```

bkamins:
Are you benchmarking against DataFrames.jl master?

The issue is that we have release-pending https://github.com/JuliaData/DataFrames.jl/blob/master/src/other/tables.jl#L5 change (it seems you are benchmarking against v20.0 release). The change was due to the fact that in really wide tables we want to avoid compilation of `Tables.columntable` (e.g. for `CSV.write`).

That is why I have opened https://github.com/JuliaData/DataFrames.jl/issues/2100 to make sure we are fast after you merge this PR and release Tables.jl (so I am waiting with fingers crossed to see it merged 😄 - thank you for all the efforts here).

bkamins:
Sorry - I see you are benchmarking it against master of DataFrames.jl. What I mean to do in https://github.com/JuliaData/DataFrames.jl/issues/2100 is to move from:
```
julia> @btime rowtable(df);
  3.230 ms (117974 allocations: 2.49 MiB)
```
to
```
julia> @btime rowtable(Tables.columntable(df));
  85.400 μs (10018 allocations: 391.42 KiB)
```
(i.e. when the user requests `rowtable` from an `AbstractDataFrame` you move through `Tables.columntable`)

What I meant earlier with my comment is that the commit did not affect the performance of `Tables.columntable`. It affects `Tables.rowtable` from `AbstractDataFrame`, but I will improve that part anyway soon with https://github.com/JuliaData/DataFrames.jl/issues/2100.

quinnj:
Merging this now, since there's a good amount of work that's built up. I still want to address https://github.com/JuliaData/Tables.jl/issues/134, and allow anyone else to review who would like before making the 1.0 release. I also want to make sure the new docs get built correctly. Please comment here or open an issue for anything else you have concerns with.

bkamins:
Tables in the documentation do not render correctly in section https://juliadata.github.io/Tables.jl/dev/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source)-1

quinnj:
Hmmm.........I tried to fix them using https://www.tablesgenerator.com/markdown_tables, but even that seems to not have them rendering. Anyone know what's going on there?

bkamins:
~~See https://github.com/JuliaData/Tables.jl/pull/138 for a fix.~~

EDIT: actually indeed something strange is happening. I will try to investigate into it. It seems that the tables that are in doc strings have the newlines removed (which in particular makes them print pad in REPL; and in general the tables are too wide for printing on 80-character terminals).

@mortenpi probably should know the right approach to use.

PR #138, opened 2020-02-08
Fix tables in markdown to improve their rendering

bkamins:


bkamins:
@quinnj - this PR seems to point to a wrong location (it targets gh-pages branch not master) so please do not merge it. I will make another PR fixing this (the tables are correct, but they are fixed in a wrong place). Sorry for messing up.

tkf:
It looks like you just need one empty line. See #143.

PR #140, opened 2020-02-08
Add Tables.isrowtable for convenience. Fixes #134

### LINKED ISSUE #134: Feature request: trait `isrowtable`
Can we have a new trait `isrowtable` to declare a certain table type implements "row table" interface? (Previous discussion: https://github.com/JuliaData/TypedTables.jl/pull/57 cc @andyferris)

The motivation is for implementing `append!(dest, src)` and alike for table `dest` with column-oriented memory layout and row-wise container semantics (e.g., TypedTables.jl, StructArrays.jl).  Ideally:

1. If `src` is a table but _not_ a container of rows (e.g., `NamedTuple` of `Vector`s), it is _not_ desirable for `append!(dest, src)` to append rows.  For example, `append!([(a=1,)], (a=[1],))` should throw.
2. If `src` is a row table but has column-oriented memory layout, it is desirable to append elements column-by-column as an _optimization_.

It is tempting to use `Tabels.columnaccess(src)` to satisfy point 2. However, `Tabels.columnaccess(src)` does not imply that `src` has rows as elements. So, it might violate point 1.

If we have `isrowtable`, we can satisfy those two points by

```julia
function append!(dest::MyContainer, src)
    if Tables.isrowtable(src) && Tables.columnaccess(src)
        return append_column_by_column!(dest, src)
    else
        return append_row_by_row!(dest, src)
    end
end
```

### Suggested specification

A type `RowTable` satisfies the row table interface if:

* `Tables.rows(x::RowTable) === x`
* `eltype(::RowTable)` is a row (not _just_ `eltype(Tables.rows(::RowTable))`)
* If `x :: RowTable` is a mutable container:
  * `push!(x, row)` must work for any _compatible_ `row`.
  * `append!(x, rows)` must work for compatible row table `rows` (i.e., `Tables.isrowtable(rows)`).
* If `x :: RowTable` is a mutable and indexable:
  * `x[i] = row` must work for any compatible `row`.

By `x` and `y` are _compatible_ I mean that row or table `x` and `y` satisfy `Set(Tables.columnnames(x)) == Set(Tables.columnnames(y))` and the types are promotable.

### Default implementation and derived interfaces

I suggest to add `isrowtable` with the default definition

```julia
isrowtable(::Type{T}) where {T} = false
isrowtable(::T) where {T} = isrowtable(T)
```

The default definition of other parts of Tables.jl can be derived from it

```julia
istable(::Type{T}) where {T} = isrowtable(T)
rowaccess(::Type{T}) where {T} = isrowtable(T)

function rows(x::T) where {T}
    isrowtable(x) && return x
    ...
end
```

This way, row tables like TypedTables.jl and StructArrays.jl can just declare

```julia
isrowtable(::Type{<:MyTable}) = true
```

and other interfaces would be derived from it.

quinnj:
cc: @tkf

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -318,3 +318,11 @@ And that's it. Our `MatrixTable` type is now a fully fledged, valid Tables.jl so
 the ecosystem. Now, this is obviously not a lot of code; but then again, the actual Tables.jl interface
 implementations tend to be fairly simple, given the other behaviors that are already defined for table types
 (i.e. table types tend to already have a `getcolumn` like function defined).
+
+### `Tables.isrowtable`
+
+One option for certain table types is to define `Tables.isrowtable` to automatically satisfy the Tables.jl interface.
+This can be convenient for "natural" table types that have row elements, along with other required properties.
+```@docs
+Tables.isrowtable
+```
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -207,7 +207,27 @@ columnaccess(::Type{<:AbstractColumns}) = true
 columns(x::AbstractColumns) = x
 schema(x::AbstractColumns) = nothing
 
-# default definitions
+"""
+    Tables.isrowtable(x) => Bool
+
+For convenience, some table objects that are naturally "row oriented" can
+define `Tables.isrowtable(::Type{TableType}) = true` to simplify satisfying the
+Tables.jl interface. Requirements for defining `isrowtable` include:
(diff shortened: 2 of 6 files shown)


tkf:
Thanks! It looks good to me. Can you also define `isrowtable(::Type{<:RowTable}) = true` etc. for existing row tables? Here is my patch against this PR: #141

PR #142, opened 2020-02-09
Idea: don't use types for documenting Row and Columns?

tkf:
I think it is a bit confusing that `Tables.Row` and `Tables.Columns` exist as actual types (though the docstrings do mention that they are only for documentation).  But maybe it's better to not introduce "dummy" types in the first place?

One idea may be to use markdown sections for documenting them. This is implemented in this PR.

---

Another way to do this is:

```julia
"""
...
"""
const Row = nothing
```

I think it is safer than `abstract type Row end` as users can't subtype `Row`.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -62,22 +62,6 @@ abstract type AbstractColumns end
 """
     Tables.Row
 
-An interface type that represents a single row of a table, with column values retrievable by name or index.
-The high-level [`Tables.rows`](@ref) function returns a `Row`-compatible
-iterator from any input table source.
-
-Any object implements the `Row` interface, by satisfying the following:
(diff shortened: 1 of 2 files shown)


src/Tables.jl
```diff
@@ -308,7 +261,7 @@ function columns end
 """
     Tables.rows(x) => Row iterator
 
-Accesses data of input table source `x` row-by-row by returning a [`Row`](@ref) iterator.
+Accesses data of input table source `x` row-by-row by returning a [`Row`](@ref Row) iterator.
```
tkf, line 264: One disadvantage of this PR is that you can't use ``` [`Row`](@ref) ```. Maybe this is fixable on Documenter.jl side, though.

quinnj (approved):
I like this; I don't mind the extra ref stuff.

PR #151, opened 2020-02-13
Make signatures of `columnindex` and `columntype` more specific

bkamins:
If I understand it correctly only `Symbol` is allowed as `name` (otherwise we get `StackOverflowError` when using name of other type)


quinnj (approved):
Thanks!

PR #162, opened 2020-03-07
lexicography (order packages in integrations)

JeffreySarnoff:


quinnj:
So the INTEGRATIONS.md file is auto-generated from the script in https://github.com/JuliaData/Tables.jl/blob/master/find_integrations.jl; would you mind making the change there? I can take a look at changing it if you don't/can't.

JeffreySarnoff:
When I run `find_integrations.jl` line-by-line in the REPL (after cd(proper dir)), the result is sorted -- why is it not in the output file here?

JeffreySarnoff:
I think changing
`pkgnames = [splitpath(x)[end] for x in pkgs]`
to
`pkgnames = sort!([splitpath(x)[end] for x in pkgs])`
would do the trick

PR #163, opened 2020-03-17
docs for INTEGRATIONS.md

Arkoniak:
Initial changes were only in INTEGRATIONS part, but it looks like atom added some modifications as well. It looks like it only removed hanging whitespaces, but if it raises any concerns I can prepare PR without these changes.


quinnj (approved):
This is great! Thanks!

PR #183, opened 2020-07-23
add `matrix(table::MatrixTable)` method

### LINKED ISSUE #182: Copying in matrix(table(A)) when A is a matrix
I was expecting that if `A` is a matrix and I do `t = table(A)` then `matrix(t)` just unwraps `A`, ie no copying.  But evidently this is not the case:

```julia
julia> A = [1 2; 2 4]
2×2 Array{Int64,2}:
 1  2
 2  4

julia> a = Tables.matrix(Tables.table(A))
2×2 Array{Int64,2}:
 1  2
 2  4

julia> A[1, 1] = 10_000
10000

julia> a
2×2 Array{Int64,2}:
 1  2
 2  4
```

Is there a reason for wanting this behaviour?

OkonSamuel:
This PR resolves #182 by adding specific method for `matrix(table::MatrixTable)`. This avoids creating a copy of the wrapped matrix.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -198,7 +198,7 @@ end
     rt = [(a=1, b=4.0, c="7"), (a=2, b=5.0, c="8"), (a=3, b=6.0, c="9")]
     nt = (a=[1,2,3], b=[4.0, 5.0, 6.0])
 
-    mat = Tables.(rt)
+    mat = Tables.matrix(rt)
     @test nt.a == mat[:, 1]
     @test size(mat) == (3, 3)
     @test eltype(mat) == Any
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -241,9 +241,12 @@ end
     @test propertynames(mattbl) == propertynames(matrow) == [:Column1, :Column2, :Column3]
 
     # #155
-    T = Tables.table(hcat([1,2,3],[1,2,3]))
+    m = hcat([1,2,3],[1,2,3])
+    T = Tables.table(m)
     M = Tables.matrix(T)
     @test M[:, 1] == [1, 2, 3]
+    # 182
+    @test M === m #checks that both are the same object in memory
```
quinnj, line 250: Mind adding a test for the `permutedims` case as well? Just as a coverage thing?

OkonSamuel: Okay

PR #187, opened 2020-08-10
Add parallel reduction supports for RowIterator and NamedTupleIterator

tkf:
This PR implements [SplittablesBase.jl](https://github.com/JuliaFolds/SplittablesBase.jl) interface `halve` on `RowIterator` and `NamedTupleIterator`.  This let us use parallel reductions built on top of SplittablesBase.jl such as [Transducers.jl](https://github.com/JuliaFolds/Transducers.jl), [ThreadsX.jl](https://github.com/tkf/ThreadsX.jl), and [FLoops.jl](https://github.com/JuliaFolds/FLoops.jl):

```julia
using Tables
table = Tables.rows((key = 1:1000, value = randn(1000)))

using FLoops
using UnPack: @unpack

@floop for row in table
    @unpack key, value = row
    @reduce() do (kmax; key), (vmax; value)
        if vmax < value
            vmax = value
            kmax = key
        end
    end
end
@show kmax vmax
```

A tricky part of this PR is that, since [`SplittablesTesting.test_ordered`](https://juliafolds.github.io/SplittablesBase.jl/dev/#SplittablesTesting.test_ordered) uses `isequal` to compare items (rows), I needed to relax `isequal` to ignore the storage type of columns.  The difference is that

```julia
isequal(
    first(Tables.rows((a = view([0], 1:1),))),
    first(Tables.rows((a = [0],))),
)
```

is `false` before this PR and `true` after this PR.  I think it makes sense that `ColumnsRow` to be compared as if they are lowered to `NamedTuple`s.  This is also compatible with that the equalities on arrays ignore the type

```julia
julia> [0] == view([0], 1:1)
true
```

What do you think?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,12 +10,14 @@ TableTraits = "3783bdb8-4a98-5b6b-af9a-565f29a5fe9c"
 IteratorInterfaceExtensions = "82899510-4779-5014-852e-03e436cf321d"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 DataValueInterfaces = "e2d170a0-9d28-54be-80f0-106bbe20a464"
+SplittablesBase = "171d559e-b47b-412a-8079-5efa626c420e"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
 QueryOperators = "2aef5ad7-51ca-5a8f-8e88-e75cf067b44b"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+SplittablesTesting = "3bda5eb5-c32a-4f64-8618-df3be8968470"
 
 [compat]
 julia = "1"
@@ -25,4 +27,4 @@ IteratorInterfaceExtensions = "0.1.1, 1"
 TableTraits = "0.4.1, 1"
 
 [targets]
-test = ["Test", "DataValues", "QueryOperators", "SparseArrays"]
+test = ["Test", "DataValues", "QueryOperators", "SparseArrays", "SplittablesTesting"]
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -1,6 +1,6 @@
 module Tables
 
-using LinearAlgebra, DataValueInterfaces, DataAPI, TableTraits, IteratorInterfaceExtensions
+using LinearAlgebra, DataValueInterfaces, DataAPI, TableTraits, IteratorInterfaceExtensions, SplittablesBase
 
 export rowtable, columntable
 
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -26,7 +26,7 @@ Base.@propagate_inbounds getcolumn(c::ColumnsRow, i::Int) = getcolumn(getcolumns
 Base.@propagate_inbounds getcolumn(c::ColumnsRow, nm::Symbol) = getcolumn(getcolumns(c), nm)[getrow(c)]
 columnnames(c::ColumnsRow) = columnnames(getcolumns(c))
 
-@generated function Base.isless(c::ColumnsRow{T}, d::ColumnsRow{T}) where {T <: NamedTuple{names}} where names
+@generated function Base.isless(c::ColumnsRow{<:NamedTuple{names}}, d::ColumnsRow{<:NamedTuple{names}}) where names
     exprs = Expr[]
     for n in names
         var1 = Expr(:., :c, QuoteNode(n))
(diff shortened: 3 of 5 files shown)


quinnj:
Can you share a bit more on the motivation here? Obviously, we want to be cautious/wary of taking on new dependencies when it will also affect so many downstream packages. Questions that pop up in my mind:
1) What are some examples of what you could do w/ the implementation here?
2) Why only implemented for `NamedTupleIterator` and not rows/columns more generally?
3) What kind of package is SplittablesBase.jl? Minimal? Lots of changes? What kind of commitment to stability there?

tkf:
Hi, thanks for the response and sorry for this late reply.

> Obviously, we want to be cautious/wary of taking on new dependencies when it will also affect so many downstream packages.

Yes, I understand this and I should've clarified what SplittablesBase.jl is.

Essentially I am hoping `halve` to be a fundamental infrastructure for parallel processing in Julia in the sense that `iterate`, at the moment, is a fundamental infrastructure for sequential processing.  The goal is to make `halve`+`iterate` (or `halve`+`foldl`) the interface between the data structures (tables, arrays, dicts, sets, strings, ...) and parallel processing functions (map, reduce, group-by, join, ...).

I probably should open an RFC in JuliaLang/julia but I've been a bit hesitant to do so since I don't feel like this interface is tested outside my packages.  I thought of this PR as a step toward accumulating such experience.

> 1. What are some examples of what you could do w/ the implementation here?

The example in the OP with FLoops.jl is one thing. We'd be able to use ThreadsX with this. I think [ThreadsX.jl + OnlineStats.jl integration](https://github.com/tkf/ThreadsX.jl#onlinestatsjl) is appealing to the Tables.jl users.  Underneath, they all boils down to [`Transducers.foldxt`](https://juliafolds.github.io/Transducers.jl/dev/reference/manual/#Transducers.foldxt) that uses `halve`.  For example, you can compute min/max of max/min over columns a, b and c by `foldxt(ProductRF(min, max), table |> Map(r -> (max(r.a, r.b, r.c), min(r.a, r.b, r.c))))` in one go (OK, I have no idea when you need this particular function but it's a fun example).

> 2. Why only implemented for `NamedTupleIterator` and not rows/columns more generally?

If it is already an array, the generic fallback in SplittablesBase.jl covers it already. So, I don't need to add a specific implementation for `RowTable`.  I can't provide an implementation for `AbstractColumns` (or column table in general) because I'd like to keep `halve` and `iterate` consistent in the sense that each `halve` implementation satisfies what I call "`vcat` law":

> > (1) If the original collection is ordered, concatenating the
> > sub-collections returned by `halve` must create a collection that is
> > equivalent to the original collection.  More precisely,
> >
> > ```julia
> > isequal(
> >     vec(collect(collection)),
> >     vcat(vec(collect(left)), vec(collect(right))),
> > )
> > ```
> >
> > must hold.
> >
> > --- https://juliafolds.github.io/SplittablesBase.jl/dev/#SplittablesBase.halve

> 3. What kind of package is SplittablesBase.jl? Minimal? Lots of changes? What kind of commitment to stability there?

My intention is making it very minimal although I have to put the implementation for `Base` there.  It currently also contains the code for testing.  However, the public API is to use it via a shim package [`SplittablesTesting`](https://juliafolds.github.io/SplittablesBase.jl/dev/#SplittablesTesting.test_ordered).  So, I can remove it at any point without introducing breaking changes.

I think it's almost 1.0-ready but there is one specification of an optional API `amount` https://github.com/JuliaFolds/SplittablesBase.jl/issues/31 that I want to clarify before 1.0.

If you want to postpone merging this at least until SplittablesBase.jl hits 1.0, I think that's a very reasonable decision.  I can extract out this PR to a separate package SplittableTables.jl for this to work (by touching the internals of Tables.jl a bit).  But it'd be nice if we can tweak `isequal` as in this PR (as this is impossible to do outside Tables.jl without a serious type-piracy).

mattwigway:
This is a really useful PR, but I see @quinnj's point about introducing an extra dependency on all of the downstream dependents of Tables. Maybe it would make more sense to have a separate package `SplittableTables.jl` that implements `halve()` for table iterators?

ParadaCarleton:
> If it is already an array, the generic fallback in SplittablesBase.jl covers it already.

I'm not sure this is true, given `halve` doesn't work with `DataFrameRows` (which is an `AbstractVector`).

> This is a really useful PR, but I see @quinnj's point about introducing an extra dependency on all of the downstream dependents of Tables. Maybe it would make more sense to have a separate package `SplittableTables.jl` that implements `halve()` for table iterators?

I think SplittablesBase.jl is fine, given it's a very small dependency.

cc @quinnj  and @MasonProtter -- being able to use JuliaFolds with Tables and DataFrames would be awesome.

ParadaCarleton:
> > 3\.
> 
> If it is already an array, the generic fallback in SplittablesBase.jl covers it already.

Although, I'm not sure this is true, given it doesn't work with DataFrames.jl.

asinghvi17:
Looks like this could be implemented as an extension, either in SplittablesBase or in Tables (preferably SplittablesBase though?).

quinnj:
Closing as old/stale; if someone wants to try the extension idea, I think that would be considered

PR #189, opened 2020-08-17
Avoid columnnames in rowcount fallback

juliohm:
Is it ok to refactor the fallback like this? I was having errors downstream because some table types did not define `columnnames`. Is it correct that any table type will have a fallback for `rows` and that we can just take the length of the rows for the count? Could we have similarly `columncount(table) = length(columns(table))`?

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -5,11 +5,7 @@
 # Turn any AbstractColumns into an AbstractRow iterator
 
 # get the number of rows in the incoming table
-function rowcount(cols)
-    names = columnnames(cols)
-    isempty(names) && return 0
-    return length(getcolumn(cols, names[1]))
-end
+rowcount(table) = length(rows(table))
 
 # a lazy row view into a AbstractColumns object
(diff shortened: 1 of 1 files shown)


quinnj:
No, this isn't going to work. Do note that `columnnames` is _definitely_ required by the result of `Tables.columns` or by each element iterated by `Tables.rows`, so if a table type isn't defining that, it isn't Tables.jl-compliant.

As noted elsewhere, this `rowcount` function is an internal function, not meant for public consumption or documented as part of the API. It can reliably be used on the result of `Tables.columns`, since it's only using properties that are required by `Tables.columns`, so if you're in a situation where you really need the row count like this, you should call `Tables.columns` on the input table and then you can safely call `Tables.rowcount(cols)`.

If you want to expand on your use-case, I'm happy to help figure out appropriate use of the Tables.jl API.

juliohm:
> If you want to expand on your use-case, I'm happy to help figure out appropriate use of the Tables.jl API.

That would be very helpful. I will refactor the codebase a bit, and will ping you when I have a more organized set of questions.

PR #194, opened 2020-09-01
Make find_integrations.jl a bit more general, update integrations.

tpapp:
1. Paths no longer hardcoded to @quinnj's computer.
2. Can be started from any directory.
3. Sort package names alphabetically.

Also update the list itself.


quinnj (approved):
Big big thanks for doing this!!

Issue #196, opened 2020-09-12
Type stability for columns of different types


rafaqz:
So these methods are intended to allow type stability in `getcolumn`:

```julia
Tables.getcolumn(table, ::Type{T}, i::Int, nm::Symbol)
```

I'm just wondering if `T` is mean to be e.g. `Float64` or `Vector{Float64}`? It's not totally clear in the docs. In my case it needs to be `Vector{Float64}` to be type-stable.

quinnj:
It should be the _element_ type, so `Float64`. We should certainly document this clearer, but the following should hold: `Tables.schema(tbl).types[i] === eltype(Tables.getcolumn(tbl, i))`, i.e. you should use the types returned from `Tables.schema` to "pass" to `Tables.getcolumn(tbl, T, i, nm)`.

Note that Tables.jl itself provides `Tables.eachcolumn(f, sch, tbl)` which can "map" over columns in a smart, type-stable way.

rafaqz:
That makes sense. I'll look into `eachcolumn`. 

But it would be good to get type stability in `getcolumn` where not all the columns are the same type, I think `getcolumn` is a more common use case than iterating over all columns with `eachcolumn`.

quinnj:
I'm not quite sure I follow this:

> But it would be good to get type stability in getcolumn where not all the columns are the same type

Could you show/explain your use-case a little more? `Tables.getcolumn` can indeed be type stable by passing the element type explicitly like `Tables.getcolumn(tbl, T, i, nm)`, though that does require the table source to implement the method and provide a type-stable way to get a column; I say that because, for example, you can't call `Tables.getcolumn` on a DataFrame and have it be type-stable because it doesn't provide that option at all.

rafaqz:
In my use case some subset of columns are lazily generated from smaller arrays to reduce unnecessary allocation in large files. Other may be backed by from various kinds of arrays I don't have control over but I want to index into directly, again to avoid allocation. So what is returned by `getcolumn` is not always the same type of object for each column.

But they might all contain `Float64`... So indexing `getcolumn` with `Float64`, `Symbol` and `Int` wont give us a specific function compiled for each column type, but for the column `eltype`. But we have different return types for some columns. So no type-stability.

Maybe this is a weird thing to do with tables idk, I could just return `Vector` instead and be done with it but this can mean a lot of memory use sometimes.

[...]

PR #197, opened 2020-09-13
column type -> column element type in docs

rafaqz:
Change most documentation referring to "column type" to "column element type". There are some remaining uses of "column type" in fallbacks.jl that do mean the type of the column object `<: AbstractVector` 

Discussed in #196


quinnj (approved):
Thanks!

PR #198, opened 2020-09-16
Remove istable for a value of type AbstractMatrix

bkamins:
I think - as in other places - we should only define a method for a type, and use a default fallback for the value.

@quinnj Was there some specific reason why in this case a method for a value was introduced?

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -1,4 +1,3 @@
-istable(::AbstractMatrix) = false
 istable(::Type{<:AbstractMatrix}) = false
 
 rows(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))


quinnj:
I don't remember really

quinnj:
Looks like there's a test that needs to be updated

bkamins:
Yes - I will fix it, but first I wanted to check the direction :smile:.

bkamins:
We should fix a bug in TrableTraits.jl and not change anything here. But let us wait till https://github.com/queryverse/TableTraits.jl/pull/21 is resolved.

devmotion:
I just came across this PR since I noticed that my implementation of `istable(::Type{<:MyArray})` was not picked up for arrays of dimension 2 when running `istable(::MyArray)`. Could this PR be merged and released?

bkamins:
After @quinnj agrees to the proposal I will update the tests.

bkamins:
@quinnj - after TableTraits.jl fix the tests pass

quinnj:
Thanks!

devmotion:
I just noticed that the PR was made against and merged into master instead of main. That seems incorrect?

quinnj:
Ah, great catch! I'll merge it over to main

PR #202, opened 2020-09-29
avoid reallocating `header`

OkonSamuel:
This PR let's the user decide whether or not they wish to reallocate `header` passed to `table` method. This is done by the introduction of `reuse_header` keyword. 
At **MLJ's** we usually pass `header::Vector{Symbol}` to the `table` method quite often and we wouldn't want that to be reallocated 

@quinnj  @bkamins thoughts?


quinnj:
How about https://github.com/JuliaData/Tables.jl/pull/203 as a simpler alternative?

OkonSamuel:
closing in favor of #203

PR #203, opened 2020-09-29
Avoid allocating header symbol vector for every matrix input

quinnj:


OkonSamuel:
thanks.

OkonSamuel:
@quinnj could you please delete 
```julia
""", but `header` is always stored as freshly allocated object."""
```
from `table`'s docstring.

src/matrix.jl
```diff
@@ -51,8 +51,8 @@ An optional keyword argument iterator `header` can be passed which will be conve
 `Vector{Symbol}` to be used as the column names. Note that no copy of the `AbstractMatrix`
 is made, but `header` is always stored as freshly allocated object.
 """
-function table(m::AbstractMatrix; header=[Symbol("Column$i") for i = 1:size(m, 2)])
-    symbol_header = [Symbol(h) for h in header]
+function table(m::AbstractMatrix; header=nothing)
```
OkonSamuel:
```suggestion
function table(m::AbstractMatrix; header=[Symbol("Column$i") for i = 1:size(m, 2)])
```
it still has to have a reasonable default

quinnj:
Updated

src/matrix.jl
quinnj: It hits this code below because `nothing` isn't `Vector{Symbol}`

OkonSamuel: Yeah. could you revert to using `header= [Symbol("Column$i") for i = 1:size(m, 2)]` or is there a specific reason for using `header=nothing`

quinnj:
Alright, pushed a cleaner update. @bkamins, there shouldn't be any issue here, right? I think it's fine because we don't expect the header to ever be changed at all.

src/matrix.jl
```diff
@@ -49,10 +49,10 @@ Wrap an `AbstractMatrix` (`Matrix`, `Adjoint`, etc.) in a `MatrixTable`, which s
 Tables.jl interface. This allows accessing the matrix via `Tables.rows` and `Tables.columns`.
 An optional keyword argument iterator `header` can be passed which will be converted to a
 `Vector{Symbol}` to be used as the column names. Note that no copy of the `AbstractMatrix`
-is made, but `header` is always stored as freshly allocated object.
+is made, but unless the `header` is a `Vector{Symbol}`, a new copy is allocated.
```
bkamins: I think it is enough to write:
```suggestion
is made.
```
since above we say that `header` is converted to `Vector{Symbol}` which implies that if it is `Vector{Symbol}` no copy is made and otherwise there is a copy (this is how `convert` works, although internally we do not use `convert` we do essentially the same).

bkamins:
The reason I originally proposed to make a copy is that the passed matrix is guaranteed not to change its dimensions (this is how matrices work), but if a header is not owned by the created object it technically can be resized (since it is a `Vector`), which will make the object have inconsistent data or its contents changed (this can potentially lead to surprising results).

However, I am OK to change this and require on the user side to either make a `copy` when passing `Vector{Symbol}` or ensure that the passed vector is not mutated later.

bkamins (approved):

Issue #207, opened 2020-10-17
Don't infer schema from the first object in an array


sa-:
This throws an error `KeyError: key :a not found`
```julia
using DataFrames, JSON3
lines = [
    """{"a": 1}""",
    """{"b": 2}""",
]
DataFrame(JSON3.read.(lines))
```

This ignores column b
```julia
lines = [
    """{"a": 1}""",
    """{"a": 2, "b": 2}""",
]
DataFrame(JSON3.read.(lines))
```

I would expect this input:
```julia
lines = [
    """{"a": 1}""",
    """{"b": 2}""",
]
DataFrame(JSON3.read.(lines))
```

to have this output: 
```
│ Row │ a      │ b      │
│     │ Union… │ Union… │
├─────┼────────┼────────┤
│ 1   │ 1      │        │
│ 2   │        │ 2      │
```

bkamins:
I have transferred the issue as it is Tables.jl not DataFrames.jl related.

sa-:
I would be happy to contribute as well if I can get some pointers on where to start

quinnj:
Yes, I've been thinking about this a fair bit recently; in an ideal world, the default `Tables.columns(non_homogenous_rows)` routine would do this kind of "unioning" by default. Unfortunately, this unioning is necessarily less performant that taking the "column names" from the first row, because on each row, we're basically limited to `[Tables.getcolumn(row, nm) for nm in Tables.columnnames(row)]` kind of pattern to get all values on each row.

I'm torn: on the one hand, this would be a much better default for non-schema inputs; on the other hand, it's painful to incur the performance hit for _all_ non-schema row tables just because of the tiny minority of these kinds of input tables w/ non-homogenous rows.

I'm going to experiment with a few things, but would be interested to hear others thoughts here.

bkamins:
I would add a kwarg handling this (just as we do in DataFrames.jl for `push!`, `append!` and `vcat`). Indeed having to check the consistency is expensive (also in DataFrames.jl).

Actually there are three natural scenarios:
1. fast: take first row
2. slow: union
3. slow: take first row BUT do a strict column names consistency check

Then the only question would be what should be the default (the fast or the safe path) - but here probably users should comment on what they prefer.

sa-:
For relatively small datasets, where coding time is much greater than computing time, an `infer_schema` kwarg that takes a performance hit would work out well.

For larger datasets there is a another fast option: let the person specify a schema themselves.

[...]

PR #212, opened 2020-11-08
Add dictcolumntable and dictrowtable functions

### LINKED ISSUE #161: Column order in a table from a Dict
This is a synchronization with DataFrames.jl issue.

See https://github.com/JuliaData/DataFrames.jl/issues/2119 for a discussion.

Essentially the question is do we want to change the behaviour from undefined order of columns to sorted columns if a passed dictionary is a `Dict`.

quinnj:
Fixes #161 and #207. This defines two table types, `DictColumnTable` and
`DictRowTable`, along wtih corresponding constructors, `dictcolumntable`
and `dictrowtable`. For #161, both table types hold their own schema,
which provides a way to specify the "ordering" of the tables, even
though the underlying data is stored in `Dict`s. For #207, both these
table types employ a "column unioning" approach for schema-less input
row iterators. This is similar to the `cols=:union` behavior you get
with DataFrames `push!`, and at least provides a way for users to get
this behavior when so desired. I like this approach because to even
implement the functionality, you basically need to treat rows/columns as
`Dict`s, to check if keys are present or not, and they're very flexible
in terms of adding additional key-value pairs (much easier than trying
to "update" `NamedTuple`s). So this is a bit of a compromise: we keep
the current great performance of `Tables.columns` fallback via
`Tables.buildcolumns` for schema-less row iterator inputs, but provide a
pretty simple "intermediate" table type that will union columns if
needed, via `dictcolumntable` and `dictrowtable`.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -204,6 +204,8 @@ Tables.partitions
 Tables.partitioner
 Tables.rowtable
 Tables.columntable
+Tables.dictrowtable
+Tables.dictcolumntable
 Tables.namedtupleiterator
 Tables.datavaluerows
 Tables.nondatavaluerows
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -534,6 +534,9 @@ include("tofromdatavalues.jl")
 # matrix integration
 include("matrix.jl")
 
+# dict tables
+include("dicts.jl")
+
 """
     Tables.columnindex(table, name::Symbol)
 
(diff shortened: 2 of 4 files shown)


ablaom:
@OkonSamuel I wonder if we might make use of these?

OkonSamuel:
> @OkonSamuel I wonder if we might make use of these?

Yes.

Issue #149, opened 2020-02-13
Column-oriented counterpart API of namedtupleiterator?


tkf:
Is it reasonable to add a counterpart of `Tables.namedtupleiterator` for column-oriented access? Something like this:

```julia
namedtupletable(table) =
    foldl(Tables.columnnames(table); init = NamedTuple()) do nt, name
        (; nt..., name => Tables.getcolumn(table, name))
    end
```

Compared to `Tables.columntable`, it is a very cheap operation for "narrow" column tables since this does not guarantee that returned columns are `Vector`s.  It makes easy to cross API boundaries when the callee does not know about Tables API.  This is also useful when dealing with untyped column tables.

This is what I needed when implementing https://github.com/JuliaArrays/StructArrays.jl/pull/117.

quinnj:
I'd be open to a PR showing what the functionality would look like. I'd also be open to making `Tables.columntable` more flexible if that would also help, instead of needing to introduce a new concept for a very similar operation.

PR #219, opened 2020-12-03
Add `namedtupleiterator`/`rowtable` docs cross-ref

nickrobinson251:
Attempts to makes `namedtupleiterator` a little more discoverable :) 

~(Aside: would it be more discoverable if renamed `rowiterator`? especially if there were a `columniterator` too? https://github.com/JuliaData/Tables.jl/issues/149.)~ Edit: no, since `RowIterator` is a different thing. Ignore that. 😆 Added xrefs to `rows` too.


quinnj (approved):
Thanks!

PR #224, opened 2020-12-10
Migrate CI to GitHub Actions

ararslan:
Travis is not currently providing credits for OSS repositories. I copied the setup from Example.jl, happy to remove any builds deemed unnecessary (or any committers can push to this branch).


quinnj (approved):
Thanks! I added the keys/secrets, so this should be good to go.

PR #226, opened 2020-12-23
improved behavior of Tables.Columns

ExpandingMan:
This PR fixes a few odd loopholes in the behavior of `Tables.Columns`.

First, we now check if the argument to `Tables.Columns` is itself a table.  It took me a little while to convince myself that this indeed must be the case, but once I did, I realized the current behavior is permissive of some pretty strange things.  For example
```julia
Tables.Columns(nothing)
```
returns a perfectly valid (albeit empty) table.  To check for this we of course have to sacrifice `Tables.Columns` being a no-op, but this seems like a pretty small sacrifice to sanitize the behavior.

Second, `Tables.Columns` did not itself have a definition for `Tables.schema`, which meant that `Tables.Columns` itself was actually broken for much of the Tables.jl interface.  Again, since it can only wrap a table, I simply take the schema of the object it wraps.  Technically `Tables.AbstractColumns` should probably have a fallback of `Tables.schema`, however, since this cannot infer the types, it seems pretty useless in practice so I left it alone.

I have added unit tests for the above, as well as some unit tests making sure matrices cooperate nicely with `Tables.Columns` (I started uncovering these issues when trying to wrap a table in `Tables.Columns` that was originally derived from a matrix.)

I've also taken the liberty of changing the name of the `Columns` struct in `runtests.jl` to `TestColumns`, as I initially found the name shadowing confusing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.2.3"
+version = "1.3.0"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.2.2"
+version = "1.2.3"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -243,7 +243,6 @@ struct Columns{T} <: AbstractColumns
     x::T
 end
 
-Columns(x) = Columns{typeof(x)}(x)
 Columns(x::AbstractColumns) = x
 
 # Columns can only wrap something that is a table, so we pass the schema through
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -241,9 +241,17 @@ to provide useful default behaviors (allows any `AbstractColumns` to be used lik
 """
 struct Columns{T} <: AbstractColumns
     x::T
+
+    function Columns{T}(x::T) where {T}
+        istable(x) ? new{T}(x) : throw(ArgumentError("Columns can only wrap an object for which `Tables.istable` is true"))
+    end
 end
 
+Columns(x) = Columns{typeof(x)}(x)
 Columns(x::AbstractColumns) = x
+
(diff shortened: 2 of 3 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.2.2"
+version = "1.2.3"
```
bkamins: I think it should be 1.3 unless @quinnj thinks that the fact that methods you add were missing was a bug.

ExpandingMan: I think it's fair to say it's a bug, since `Tables.Columns` seemed to promise a table interface but was broken due to this issue.  I don't have a strong opinion though, if you want 1.3 let me know and I'll change it.

bkamins: I would leave it for @quinnj to decide 😄.

nalimilan: I'd be inclined to use a more restrictive definition of "bugfix". Patch releases should only include changes that really fix errors or invalid (misleading/dangerous) results. Changes in behaviors (such as the one here) should definitely be avoided as much as possible as they may break code. People should be able to update to patch releases without fearing breakage.

ExpandingMan: I've increased the version number.

ExpandingMan:
bump

src/Tables.jl
```diff
@@ -241,10 +241,18 @@ to provide useful default behaviors (allows any `AbstractColumns` to be used lik
 """
 struct Columns{T} <: AbstractColumns
     x::T
+
+    function Columns{T}(x::T) where {T}
+        istable(x) ? new{T}(x) : throw(ArgumentError("Columns can only wrap an object for which `Tables.istable` is true"))
```
quinnj: Hmmm, I don't think I like this restriction. In general, we discourage use of `Tables.istable` and encourage instead to just call `Tables.rows` or `Tables.columns` as needed and code will error if the input isn't valid.

We probably need more docs/usage examples around how to make sure you do `Tables.Column(Tables.columns(x))` to ensure everything works properly. @nalimilan and I recently had a lengthy discussion of adding some additional convenience methods around all this, I'll have to see if I can find it and put the results of our discussion up in an issue.

ExpandingMan: Like I said, the current behavior can lead to some pretty weird stuff.  Regardless, I'd be happy to take this out if we can get it merged, it could always be changed later.

ExpandingMan:
This PR no longer checks if the argument passed to `Tables.Columns` is a table.  In my opinion, we at the very least need drastically clearer documentation.  As I've pointed out, currently you can get some pretty strange behavior out of `Tables.Columns` (in particular every invalid table becomes an empty table).

In the meantime, hopefully we can merge this PR and tag if there are no more objections (it's blocking a package release for me).

Thanks all.

quinnj (approved):
Sorry for the delay here and thanks for the test updates, those look great.

PR #227, opened 2021-01-02
Update INTEGRATIONS.md

StatisticalMice:
Re-created INTEGRATIONS.md by running the find_integrations.jl script.

There are 29 new packages with direct dependency to Tables.jl. TypedTables has moved. DimensionalData did not have the link in the same format as the other lines.


quinnj (approved):
Thanks!

PR #233, opened 2021-03-09
add AbstractRow equivalent of Base.merge

jrevels:
This tiny utility function kept coming up in Beacon's usage of (self-authored) Tables.jl-compliant interfaces, where functions on row-like values are duck-typed to accept any `AbstractRow`-compliant value. When calling such functions, I find I often want to concisely transform the row at the callsite to patch in a slightly modified field (or set of fields). However, as a caller, I'm also usually manipulating whatever `AbstractRow`-compliant values are passed to me, so I need to ensure my manipulations are sufficiently generic.

Thus, I end up using this little utility in a variety of places. Is there interest in having it live here? If not, I'm sure I can find a different place to put it 😁

If it's a good idea for this to live here, feel free to bikeshed the ridiculously vague name of the function...I have also called it `from` in the past, which isn't any better 😛. I originally called it `mergerows` (after `Base.merge`) but:

1.  `mergerows` sounds like it might be referring to a row-wise reduction on a table (which this is not)
2. I often use it in argument position (e.g. `myfunction(with(row; field=f(row.old)))`), so the conciseness is nice
3. The kwarg form is very convenient, and `Base.merge` doesn't have that form, reducing the resemblance a tad

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.3.2"
+version = "1.3.3"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.3.3"
+version = "1.4.0"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -127,3 +127,22 @@ end
     end
     return
 end
+
+"""
+    with(row; patch...)
+    with(row, patches...)
+
+Return a `NamedTuple` by merging `row` (a `AbstractRow`-compliant value) with `patches` (one or more
+`AbstractRow`-compliant values) via `Base.merge`. In other words, this function is similar to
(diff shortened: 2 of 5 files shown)


src/utils.jl
```diff
@@ -127,3 +127,22 @@ end
...
+
+Return a `NamedTuple` by merging `row` (a `AbstractRow`-compliant value) with `patches` (one or more
+`AbstractRow`-compliant values) via `Base.merge`. In other words, this function is similar to
+`Base.merge(::NamedTuple, ::NamedTuple...)`, but accepts `AbstractRow`-compliant values intead of
+`NamedTuple`s.
+
+A convenience method `with(row; patch...) = with(row, patch)` is defined that enables `patch`
+fields to be specified as keyword arguments.
+"""
+with(row; patch...) = with(row, patch)
```
quinnj:
```suggestion
with(row; patch...) = with(row, patch.data)
```
this will be more efficient as it skips wrapping a NamedTuple in a Row and then constructing a NamedTuple from that

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.3.2"
+version = "1.3.3"
```
quinnj:
```suggestion
version = "1.4.0"
```
new features are just a minor bump

quinnj (commented):
Yeah, I'm ok w/ this living here. I actually had a use-case for this just the other day. How about calling it `rowmerge`?

nalimilan:
`rowmerge` sounds definitely better than `with`. :-)

What's the policy regarding adding features to Tables.jl? Shouldn't this live in TableOperations instead?

quinnj:
This isn't really a "table" operation though; just a row merge. We have a few other row utilities like `NamedTuple(::AbstractRow)`, and this isn't much code, so I'm fine w/ having it here.

jrevels:
Settled on `rowmerge` via Slack convo. This PR should now be ready to go.

Also added a `.gitignore` as I kept accidentally checking in `Manifest.toml` otherwise lol.

quinnj (approved):

PR #234, opened 2021-03-17
Fix slow performance in Tables.columns(::RowTable)

jacobadenbaum:
When a `RowTable` has relatively few fields, the fallback `Tables.columns` constructor works very well, but it has really bad performance on wide tables (more than ~6 fields in each row)

```julia
using Tables, BenchmarkTools
# 3 fields
tb = [(a = randn(), b = randn(), c = randn()) 
      for i in 1:100_000]
@benchmark Tables.columns($tb)    

# BenchmarkTools.Trial: 
#   memory estimate:  2.29 MiB
#   allocs estimate:  6
#   --------------
#   minimum time:     143.391 μs (0.00% GC)
#   median time:      173.899 μs (0.00% GC)
#   mean time:        204.647 μs (5.09% GC)
#   maximum time:     3.898 ms (94.06% GC)
#   --------------
#   samples:          10000
#   evals/sample:     1


# 5 fields
tb = [(a = randn(), b = randn(), c = randn(), d = randn(), e = randn()) 
      for i in 1:100_000]
@benchmark Tables.columns($tb)    

# BenchmarkTools.Trial: 
#   memory estimate:  3.82 MiB
#   allocs estimate:  10
#   --------------
#   minimum time:     305.083 μs (0.00% GC)
#   median time:      414.461 μs (0.00% GC)
#   mean time:        450.670 μs (5.81% GC)
#   maximum time:     4.582 ms (89.16% GC)
#   --------------
#   samples:          10000
#   evals/sample:     1

# 6 fields
tb = [(a = randn(), b = randn(), c = randn(), d = randn(), e = randn(), f = randn()) 
      for i in 1:100_000]
@benchmark Tables.columns($tb)  

# BenchmarkTools.Trial: 
#   memory estimate:  59.46 MiB
#   allocs estimate:  2096442
#   --------------
#   minimum time:     60.605 ms (2.27% GC)
#   median time:      65.735 ms (2.36% GC)
#   mean time:        68.807 ms (5.05% GC)
#   maximum time:     136.108 ms (44.86% GC)
#   --------------
#   samples:          73
#   evals/sample:     1
```
This appears to be a due to a failure of type inference on the names of the entries in the `RowTable`, and causes problems when doing things like constructing a DataFrame from simulation results that return vectors of NamedTuples.   

Since the Tables interface for vectors of NamedTuples is defined here in the base package, I'm pretty sure that this needs to be fixed here, as opposed to downstream.  

# This PR
Implements a custom `Tables.columns` method for `RowTable` that generates fast code for `RowTable` objects with arbitrarily many fields.  Unfortunately, the only way I've figured out how to do this is with generated functions that manually unroll the loop over the fields of the named tuples in the `RowTable`.  

After this PR, performance remains the same for narrow tables, but is substantially improved for wider tables (in this simple example, we get a 150x speedup for a table with 6 columns)
```julia
using Tables, BenchmarkTools
# 3 fields
tb = [(a = randn(), b = randn(), c = randn()) for i in 1:100_000]
@benchmark Tables.columns($tb)  

# BenchmarkTools.Trial: 
#   memory estimate:  2.29 MiB
#   allocs estimate:  6
#   --------------
#   minimum time:     140.169 μs (0.00% GC)
#   median time:      212.595 μs (0.00% GC)
#   mean time:        259.682 μs (9.79% GC)
#   maximum time:     6.743 ms (94.79% GC)
#   --------------
#   samples:          10000
#   evals/sample:     1  

# 5 fields
tb = [(a = randn(), b = randn(), c = randn(), d = randn(), e = randn()) for i in 1:100_000]
@benchmark Tables.columns($tb)    # 

# BenchmarkTools.Trial: 
#   memory estimate:  3.82 MiB
#   allocs estimate:  10
#   --------------
#   minimum time:     300.329 μs (0.00% GC)
#   median time:      424.496 μs (0.00% GC)
#   mean time:        510.582 μs (9.07% GC)
#   maximum time:     5.376 ms (89.58% GC)
#   --------------
#   samples:          9777
#   evals/sample:     1

# 6 fields
tb = [(a = randn(), b = randn(), c = randn(), d = randn(), e = randn(), f = randn()) for i in 1:100_000]
@benchmark Tables.columns($tb)    

# BenchmarkTools.Trial: 
#   memory estimate:  4.58 MiB
#   allocs estimate:  12
#   --------------
#   minimum time:     397.488 μs (0.00% GC)
#   median time:      532.182 μs (0.00% GC)
#   mean time:        573.441 μs (8.48% GC)
#   maximum time:     6.427 ms (89.16% GC)
#   --------------
#   samples:          8710
#   evals/sample:     1
```

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -109,7 +109,7 @@ allocatecolumn(T, len) = DataAPI.defaultarray(T, 1)(undef, len)
 @inline function _allocatecolumns(::Schema{names, types}, len) where {names, types}
     if @generated
         vals = Tuple(:(allocatecolumn($(fieldtype(types, i)), len)) for i = 1:fieldcount(types))
-        return :(NamedTuple{Base.map(Symbol, names)}(($(vals...),)))
+        return :(NamedTuple{$(Base.map(Symbol, names))}(($(vals...),)))
     else
         return NamedTuple{Base.map(Symbol, names)}(Tuple(allocatecolumn(fieldtype(types, i), len) for i = 1:fieldcount(types)))
     end
@@ -131,17 +131,10 @@ end
     L = Base.IteratorSize(T)
     len = Base.haslength(T) ? length(rowitr) : 0
     nt = allocatecolumns(schema, len)
-    buildcolumns!(nt, schema, rowitr)
-    
-    return nt
-end
-
-function buildcolumns!(nt, schema, rowitr::T) where {T}
-    L = Base.IteratorSize(T)
-    len = Base.haslength(T) ? length(rowitr) : 0
     for (i, row) in enumerate(rowitr)
         eachcolumns(add!, schema, row, nt, L, i)
     end
+    return nt
 end
 
 @inline add!(dest::AbstractArray, val, ::Union{Base.HasLength, Base.HasShape}, row) = setindex!(dest, val, row)
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -131,10 +131,17 @@ end
     L = Base.IteratorSize(T)
     len = Base.haslength(T) ? length(rowitr) : 0
     nt = allocatecolumns(schema, len)
+    buildcolumns!(nt, schema, rowitr)
+    
+    return nt
+end
+
+function buildcolumns!(nt, schema, rowitr::T) where {T}
+    L = Base.IteratorSize(T)
+    len = Base.haslength(T) ? length(rowitr) : 0
     for (i, row) in enumerate(rowitr)
         eachcolumns(add!, schema, row, nt, L, i)
     end
-    return nt
 end
 
 @inline add!(dest::AbstractArray, val, ::Union{Base.HasLength, Base.HasShape}, row) = setindex!(dest, val, row)
--- a/src/namedtuples.jl
+++ b/src/namedtuples.jl
@@ -79,6 +79,40 @@ function Base.iterate(rows::NamedTupleIterator{Nothing}, state::Tuple{Val{names}
     return NamedTuple{names}(Tuple(getcolumn(row, nm) for nm in names)), (Val(names), (st,))
 end
 
+# Faster columns implementation
+@generated function columns(rows::RowTable)
+    T = eltype(rows)
+    # Handle the empty non-concrete named tuple case with a fallback to the
+    # generic buildcolumns
+    !isconcretetype(T) && return quote
+        CopiedColumns(buildcolumns(schema(rows), rows))
+    end
+
+    # Generate expression for named tuple of columns, entirely unfilled
+    exs = map(enumerate(fieldnames(T))) do (i, field)
+        return :($field = begin
+            TT = fieldtype($T, $i)
+            Vector{TT}(undef, length(rows))
+        end)
+    end
+    col_expr = Expr(:tuple, exs...)
+
+    # Generate expression for unrolled loop over the fieldnames
+    inner = Expr(:block)
+    for field in fieldnames(T)
+        ex = :(cols.$field[i] = rows[i].$field)
+        push!(inner.args, ex)
+    end
+
+    return quote
+        cols = $col_expr
+        for i in eachindex(rows)
+            $inner
+        end
+        return CopiedColumns(cols)
+    end
+end
+
 # sink function
 """
     Tables.rowtable(x) => Vector{NamedTuple}
--- a/src/namedtuples.jl
+++ b/src/namedtuples.jl
@@ -79,40 +79,6 @@ function Base.iterate(rows::NamedTupleIterator{Nothing}, state::Tuple{Val{names}
     return NamedTuple{names}(Tuple(getcolumn(row, nm) for nm in names)), (Val(names), (st,))
 end
 
-# Faster columns implementation
-@generated function columns(rows::RowTable)
-    T = eltype(rows)
-    # Handle the empty non-concrete named tuple case with a fallback to the
-    # generic buildcolumns
-    !isconcretetype(T) && return quote
-        CopiedColumns(buildcolumns(schema(rows), rows))
-    end
-
-    # Generate expression for named tuple of columns, entirely unfilled
-    exs = map(enumerate(fieldnames(T))) do (i, field)
-        return :($field = begin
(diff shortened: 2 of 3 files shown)


jacobadenbaum:
Note: I think what's going on here with the decrease in project codecov is that a lot of the tests for the fallback methods are now hitting this new specialized method. I guess I can fix this in the tests by copying the existing tests for `Tables.columns` and forcing them to run on a `Vector{NamedTuple}` as well, which will force it to use the actual fallback implementation.  

I don't know if it makes sense to do that, or if there's a better way.

quinnj:
Very interesting stuff. Thanks for the context and proposed solution. It's certainly curious with the huge performance hit going from 5 -> 6 fields. I'm personally interested in digging into what `@code_typed` produces and if there's anything we could do to avoid the hit.

I like this solution though because it's decently simple. My one worry is we've taken care to avoid really really bad compilation times that can sometimes happen with generated functions. In fact, in https://github.com/JuliaData/Tables.jl/commit/a844d30a291a8c2a57c0c0a359abdc32d463b992, we introduced the notion of a `SPECIALIZATION_THRESHOLD` where we won't generate specialized code if there are too many columns. This would be manifest in this PR if you had a `Vector{T}` where `T` was a NamedTuple with hundreds or thousands of fields. In those cases, just the cost to compile the function would probably swamp any practical use of specialized code. Through benchmarking, you can typically find at what point specializing becomes prohibitive.

jacobadenbaum:
Okay, so here's the issue.  For some reason, the return type of `_allocatecolumns` isn't getting properly inferred.
```julia
@generated randtp(::Val{N}) where N = begin
    ex = Expr(:tuple)
         for i in 1:N
             name = Symbol(:x, i)
             push!(ex.args, :($name = randn()))
         end
    return ex
end
randtp(x::Int) = randtp(Val(x))
tb = [randtp(10) for i in 1:100]
@code_warntype Tables._allocatecolumns(Tables.schema(tb), tb)
```
Screen Shot 2021-03-17 at 8 06 02 PM

I've played around with it and it looks like it can be fixed with a function barrier inside `buildcolumns` to insulate the actual loop over the rows from the type instability.  

Now, the benchmark for the 6 field case looks good, without needing to use any generated functions!
```julia
tb = [(a = randn(), b = randn(), c = randn(), d = randn(), e = randn(), f = randn()) for i in 1:100_000]
@benchmark Tables.columns($tb) 
# BenchmarkTools.Trial: 
#  memory estimate:  4.58 MiB
#  allocs estimate:  19
#  --------------
#  minimum time:     403.249 μs (0.00% GC)
#  median time:      530.350 μs (0.00% GC)
#  mean time:        565.150 μs (7.66% GC)
#  maximum time:     4.124 ms (85.57% GC)
#  --------------
#  samples:          8831
#  evals/sample:     1
```

jacobadenbaum:
I mean, the ideal thing to do would be to fix the type instability in `_allocatecolumns`, but tbh I'm not sure how to do that, since it seems to have something to do with the compiler not being able to infer the results from `Base.map(Symbol, names)` 🤷

quinnj:
I replaced the `Base.map(Symbol, names)` with just `names` and that seems to fix it and all tests pass. So I think that's probably the simplest solution? I think I remember those map calls were there for `Tuple` rows, but this is only in the case where you have a known schema; and I'm not aware of any known-schema tables that produce tuple rows. And in addition, we actually enforce the names passed to `Tables.Schema` to be `Symbols` (see Tabes.jl line 412-413), so I really don't think the map calls are needed, i.e. I think it's guaranteed that the `names` of a Schema will always be Symbols.

jacobadenbaum:
Ahh, I had assumed that this was to handle some weird corner case where they got passed as strings or something.  This seems like the best fix then.  I'm glad the fix was so easy in the end.

jacobadenbaum:
Wait hang on, aren't they enforced to be either `Symbols` or `Ints`?  

https://github.com/JuliaData/Tables.jl/blob/94bbd28ff97e45587a888bf53e02f6113d8a86cf/src/Tables.jl#L409-L413

quinnj:
Ah, you're right; I remember now, we allow Tuple rows to pass in their `propertynames` as Ints and then those can be used in `Tables.getcolumn`. So how about we make this change in `Tables._allocatecolumns`:

```julia
if @generated
        vals = Tuple(:(allocatecolumn($(fieldtype(types, i)), len)) for i = 1:fieldcount(types))
        return :(NamedTuple{$(Base.map(Symbol, names))}(($(vals...),)))
    else
        return NamedTuple{Base.map(Symbol, names)}(Tuple(allocatecolumn(fieldtype(types, i), len) for i = 1:fieldcount(types)))
    end
```
i.e. just do the `Base.map(Symbol, names)` at generation time instead of runtime? That seems to also work.

jacobadenbaum:
Yeah that looks like it works well.  I've been playing around with how much one get get out of the function barrier for the tables that are above the specialization threshold (even once this change is made), and it looks as though there is some additional gain for wide (but not enormously wide) tables from 100 to ~300 fields.  But of course, once you're up in the realm of vectors of named tuples with 100 fields, you're probably doing something wrong and should reconsider your approach anyway...

jacobadenbaum:
I have updated the PR to reflect this discussion, so now it is just the one-line fix to `_allocatecolumns`

bkamins:
> But of course, once you're up in the realm of vectors of named tuples with 100 fields

I would say that having hundreds or even thousands of columns is quite usual (at least in my field), but that is where I use DataFrames.jl not `NamedTuple`s.

jacobadenbaum:
> > But of course, once you're up in the realm of vectors of named tuples with 100 fields
> 
> I would say that having hundreds or even thousands of columns is quite usual (at least in my field), but that is where I use DataFrames.jl not `NamedTuple`s.

yeah exactly!  I often deal with this as well, but the compiler basically gives up once you hit named tuples with 500 fields, so anyone who is doing this with NamedTuples is going to have bigger problems than the `Tables.columns` constructor being slow.

quinnj (approved):
Thanks again for all the help here @jacobadenbaum!

JeffreySarnoff:
very good sleuthing y'all

PR #235, opened 2021-03-26
explicitly require 1-based indexing

bkamins:
This assumption is implicit in the docstring but I think it is fundamentally important, so if we assume this I would make it explicit.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -13,7 +13,7 @@ end
 
 An interface type defined as an ordered set of columns that support
 retrieval of individual columns by name or index. A retrieved column
-must be an indexable collection with known length, i.e. an object
+must be a 1-based indexable collection with known length, i.e. an object
 that supports `length(col)` and `col[i]` for any `i = 1:length(col)`.
 `Tables.columns` must return an object that satisfies the `Tables.AbstractColumns` interface.
 While `Tables.AbstractColumns` is an abstract type that custom "columns" types may subtype for
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -343,7 +343,7 @@ materializer(::Type{T}) where {T} = columntable
 
 Accesses data of input table source `x` by returning an [`AbstractColumns`](@ref)-compatible
 object, which allows retrieving entire columns by name or index. A retrieved column
-is a 1-based object that is indexable and has a known length, i.e. supports
+is a 1-based indexable object that has a known length, i.e. supports
 `length(col)` and `col[i]` for any `i = 1:length(col)`. Note that
 even if the input table source is row-oriented by nature, an efficient generic
 definition of `Tables.columns` is defined in Tables.jl to build a `AbstractColumns`-
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -343,7 +343,7 @@ materializer(::Type{T}) where {T} = columntable
 
 Accesses data of input table source `x` by returning an [`AbstractColumns`](@ref)-compatible
 object, which allows retrieving entire columns by name or index. A retrieved column
-is an object that is indexable and has a known length, i.e. supports
+is a 1-based object that is indexable and has a known length, i.e. supports
 `length(col)` and `col[i]` for any `i = 1:length(col)`. Note that
 even if the input table source is row-oriented by nature, an efficient generic
 definition of `Tables.columns` is defined in Tables.jl to build a `AbstractColumns`-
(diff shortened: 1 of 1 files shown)


bkamins:
@quinnj - I am also not clear about the method:
```
Tables.getcolumn(::AbstractColumns, T, i::Int, nm::Symbol)
```
as in it `i` and `nm` seem to be redundant, and in the default implementation `nm` is just ignored, so why is it needed?

bkamins:
Also it seems that there was a problem with `haskey` implementation (but please double check @quinnj).

bkamins:
Finally - maybe you can consider adding `Tables.columnindex` method to `AbstractColumns` for consistency?

src/Tables.jl
```diff
@@ -179,7 +179,7 @@ Base.propertynames(r::RorC) = columnnames(r)
 Base.keys(r::RorC) = columnnames(r)
 Base.values(r::RorC) = collect(r)
 Base.haskey(r::RorC, key::Symbol) = key in columnnames(r)
-Base.haskey(r::RorC, i::Int) = 0 < i < length(columnnames(r))
+Base.haskey(r::RorC, i::Int) = 0 < i <= length(columnnames(r))
```
quinnj, line 182: Good catch!

quinnj:
> @quinnj - I am also not clear about the method:
> 
> ```
> Tables.getcolumn(::AbstractColumns, T, i::Int, nm::Symbol)
> ```
> 
> as in it `i` and `nm` seem to be redundant, and in the default implementation `nm` is just ignored, so why is it needed?

It's provided for consistency with `Tables.getcolumn(::AbstractRow, T, i::Int, nm::Symbol)` and to allow either (`AbstractColumns` or `AbstractRow`) to use `nm` _if needed_. It's easier for some implementations to key off `nm` instead of `i`, so we provide both in a "fully qualified" way for whichever is easier.

quinnj:
> Finally - maybe you can consider adding `Tables.columnindex` method to `AbstractColumns` for consistency?

Yeah, that sounds fine to me. Just an oversight if it doesn't already work.

bkamins:
>  so we provide both in a "fully qualified" way for whichever is easier.

The problem is that passed `i` and `nm` can be inconsistent. Then some methods use `i` and some `nm`. Maybe it would be good to note that this is unsafe and the user is responsible to make sure they are consistent?

quinnj:
> > so we provide both in a "fully qualified" way for whichever is easier.
> 
> The problem is that passed `i` and `nm` can be inconsistent. Then some methods use `i` and some `nm`. Maybe it would be good to note that this is unsafe and the user is responsible to make sure they are consistent?

Can you clarify what you mean? They shouldn't be inconsistent. Are you aware of cases where they are inconsistent?

bkamins:
See e.g.:
```
julia> ntv = (a=[1,2,3], b=[2,3,4])
(a = [1, 2, 3], b = [2, 3, 4])

julia> Tables.getcolumn(ntv, Int, 2, :a)
3-element Vector{Int64}:
 2
 3
 4

julia> Tables.getcolumn(DataFrame(ntv), Int, 2, :a)
3-element Vector{Int64}:
 1
 2
 3
```

and you see that since `i` and `nm` are inconsistent depending on the type of table you pass you get a different column.

quinnj:
Ok, I see what you mean; yeah, there's no validation done to ensure `i` and `nm` are consistent. Partly because implementations can overload the definition themselves, so if they're worried, they could do the consistency check themselves. Fortunately, end-users almost never call `Tables.getcolumn` directly, but rely on sink packages who are using the Tables.jl APIs. Even then, most sink packages will use `Tables.Columns` or `Tables.eachcolumn` instead of calling `Tables.getcolumn` directly, so I think the actual surface area of places people might get this wrong is pretty small.

Then again, I think this would just be a bug for whoever the `Tables.getcolumn` caller was; if you're not passing the correct indices/names, then you should fix it. We actually had a bug for this at one point with `TableOperations.select` because it produces a "subset" of the original table, but must ensure the `i` and `nm` translate from the new subset to the original parent, because it selects lazily.

Anyway, I don't think it's anything to worry about, other than perhaps making a mention in the `Tables.getcolumn` doc that callers should ensure consistency.

bkamins:
OK

src/Tables.jl
```diff
@@ -343,7 +343,7 @@ materializer(::Type{T}) where {T} = columntable
 
 Accesses data of input table source `x` by returning an [`AbstractColumns`](@ref)-compatible
 object, which allows retrieving entire columns by name or index. A retrieved column
-is an object that is indexable and has a known length, i.e. supports
+is a 1-based object that is indexable and has a known length, i.e. supports
```
```suggestion
is a 1-based indexable object that has a known length, i.e. supports
```

quinnj (approved):

PR #236, opened 2021-04-07
add shorter show method for `AbstractColumns`

ericphanson:
For example,
```julia
julia> big_tbl = (; a = [rand(100) for _ = 1:200], b = [rand(100) for _ = 1:200], c = [rand(100) for _ = 1:200]);

julia> tbl = Arrow.Table(Arrow.tobuffer(big_tbl))
Arrow.Table with 200 rows, 3 columns, and schema:
 :a  Vector{Float64} (alias for Array{Float64, 1})
 :b  Vector{Float64} (alias for Array{Float64, 1})
 :c  Vector{Float64} (alias for Array{Float64, 1})
```
This table used to print as 1.2M characters! (The “alias” printing is removed with https://github.com/JuliaLang/julia/pull/39256 I believe).

I tried to be pretty safe here since `AbstractColumns` are very general, and I'm not sure if we can expect columns to have a `length` for example.

I decided to print the schema instead of a manual column printing as in https://github.com/JuliaData/Arrow.jl/issues/168, since we don't have the `eltype` info in general (`columntypes` just asks the schema, so the fallback method only works when we have a schema), and I think maybe if Tables doesn't have a schema that can be useful info to pass along to the user (perhaps they can add one and improve performance in some cases).

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -190,7 +190,7 @@ function Base.NamedTuple(r::RorC)
     return NamedTuple{Tuple(Base.map(Symbol, names))}(Tuple(getcolumn(r, nm) for nm in names))
 end
 
-function Base.show(io::IO, x::T) where {T <: RorC}
+function Base.show(io::IO, x::T) where {T <: AbstractRow}
     if get(io, :compact, false) || get(io, :limit, false)
         print(io, "$T: ")
         show(io, NamedTuple(x))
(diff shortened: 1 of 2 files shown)


src/Tables.jl
```diff
@@ -202,6 +202,25 @@ function Base.show(io::IO, x::T) where {T <: RorC}
     end
 end
 
+function Base.show(io::IO, table::AbstractColumns; max_cols = 20)
+    nrows = try
+        string(length(Tables.getcolumn(table, 1)))
+    catch e
+        e isa MethodError || rethrow()
```
ericphanson: trying to catch a `MethodError` here if `length` isn't supported, but I wonder if there's a better way to do this

bkamins: maybe use `IteratorSize` trait? I am not sure if this would be better.

quinnj: We can assume columns have length; they have to be 1-indexable.

quinnj: We also have `Tables.rowcount` that already does this

ericphanson: ah great, thanks!

quinnj (approved):
Thanks @ericphanson !

PR #239, opened 2021-05-22
Fix Tables.schema on Tables.DictRowTable

quinnj:
Yuck; these kinds of bugs are the worst. Poor original coding (shame on
you past self), and tests that _happened_ to work by coincidence. The
problem here is that by just calling `values(getfield(x, :types))`,
we're just retrieving the types by however `Dict` internals happen to
order them, but in most cases, will _not_ match the order of column
names. We had a test for a simple case like `@test Tables.schema(drt) ==
Tables.Schema((:a, :b, :c), (Int, Float64, String))` that I guess just
happens to maintain the right ordering, even across julia versions. Oh
well, I'm guess that means not a lot of people rely on the exact schema
produced by `Tables.dictrowtable`, if they're even aware it exists and
use it. We probably need to better advertise its functionality.

### DIFF
--- a/src/dicts.jl
+++ b/src/dicts.jl
@@ -101,7 +101,7 @@ struct DictRowTable
 end
 
 isrowtable(::Type{DictRowTable}) = true
-schema(x::DictRowTable) = Schema(getfield(x, :names), values(getfield(x, :types)))
+schema(x::DictRowTable) = Schema(getfield(x, :names), [getfield(x, :types)[nm] for nm in getfield(x, :names)])
 
 struct DictRow <: AbstractRow
     names::Vector{Symbol}
--- a/src/dicts.jl
+++ b/src/dicts.jl
@@ -145,7 +145,7 @@ function dictrowtable(x)
     r = rows(x)
     L = Base.IteratorSize(typeof(r))
     out = Vector{Dict{Symbol, Any}}(undef, Base.haslength(r) ? length(r) : 0)
-    for (i, drow) in enumerate(x)
+    for (i, drow) in enumerate(r)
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -688,6 +688,7 @@ end
     ]
 
     drt = Tables.dictrowtable(rt)
+    @test Tables.schema(drt) == Tables.Schema((:a, :b, :c, :d), (Union{Missing, Int64}, Union{Missing, Float64, Int64}, Union{Missing, Int64}, Union{Missing, Int64}))
```
bkamins:
```suggestion
    @test Tables.schema(drt) == Tables.Schema((:a, :b, :c, :d), (Union{Missing, Int}, Union{Missing, Float64, Int}, Union{Missing, Int}, Union{Missing, Int}))
```

bkamins: to make tests pass on 32 bit architectures

bkamins (approved):

src/dicts.jl
```diff
@@ -145,7 +145,7 @@ function dictrowtable(x)
     r = rows(x)
     L = Base.IteratorSize(typeof(r))
     out = Vector{Dict{Symbol, Any}}(undef, Base.haslength(r) ? length(r) : 0)
-    for (i, drow) in enumerate(x)
+    for (i, drow) in enumerate(r)
```
bkamins, line 148: maybe add a test for this? I guess:
```
t = (a=(1,2,3), b=(4,5,6))
Tables.dictrowtable(t)
```
should catch the old bug. Right?

PR #241, opened 2021-06-18
Allow stored names/types in Schema for very large schemas

quinnj:
There have been a few cases of extremely wide tables where users have
run into fundamental compiler limits for lengths of tuples (as discussed
with core devs). One example is
https://github.com/JuliaData/CSV.jl/issues/635. This PR proposes for
very large schemas (> 65,000 columns), to store names/types in `Vector`
instead of tuples with the aim to avoid breaking the runtime. The aim
here is to be as non-disruptive as possible, hence the very high
threshold for switching over to store names/types. Another goal is that
downstream packages don't break with just these changes in place. I'm
not aware of any packages testing such wide tables, but in my own
testing, I've seen issues where packages are relying on the
`Tables.Schema` type parameters for names/types. There's also an issue
in DataFrames where `Tables.schema` attempts to construct a
`Tables.Schema` directly instead of using the `Tables.Schema(names,
types)` constructor. So while this PR is needed, we'll need to play
whack-a-mole with downstream packages to ensure these really wide tables
can be properly supported end-to-end. Going through those downstream
package changes, we should probably make notes of how we can clarify
Tables.jl interface docs to hopefully help future implementors do so
properly and avoid the same pitfalls.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -187,7 +187,7 @@ Base.isempty(r::RorC) = length(r) == 0
 
 function Base.NamedTuple(r::RorC)
     names = columnnames(r)
-    return NamedTuple{Tuple(Base.map(Symbol, names))}(Tuple(getcolumn(r, nm) for nm in names))
+    return NamedTuple{Tuple(map(Symbol, names))}(Tuple(getcolumn(r, nm) for nm in names))
 end
 
 function Base.show(io::IO, x::T) where {T <: AbstractRow}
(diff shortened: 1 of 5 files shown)


bkamins:
Do we test against empty schema?

ablaom:
Great, this may help https://github.com/alan-turing-institute/MLJBase.jl/issues/428

quinnj:
> Do we test against empty schema?

Looks like we don't have any explicit tests; I'll add a few.

> Great, this may help alan-turing-institute/MLJBase.jl#428

I don't think this will affect the case mentioned in that issue specifically. The core issue here is that Julia's core `Tuple`/`NamedTuple` types hit fundamental issues when larger than 67,000 elements. Probably something about overflowing memory pages or something low-level like that. The fix I'm proposing in this PR is to avoid storing column names/types in `Tuple` and `Tuple{}` types when there are more columns than 67,000, and instead store them in `Vector{Symbol}` and `Vector{Type}`. I did add the ability to call `Tables.Schema(names, types; stored=true)`, which would allow custom table types to _always_ store names/types in `Vector`s if desired.

But none of that changes that when you try to do `Tables.columntable(x)` on a really wide table that it will be slow, because it's trying to generate a `NamedTuple{names, types}` object, and that uses `Tuple`/`Tuple{}`, so it's going to be slow for # of columns larger than 10K, and will completely break when > 67K. Which reminds me that I'm going to add a note in the docs for `Tables.columntable` about that limitation. These kinds of slowdowns/limitations also apply to TypedTables and JuliaDB, and is a key driving design for DataFrames.jl storing columns as `Vector{AbstractVector}`.

quinnj:
Ok, I've made `Tables.columntable`/`Tables.rowtable`/`Tables.namedtupleiterator` all throw `ArgumentError` when the input tables are too wide; that's at least better than showing weird compiler breakage. I also tested that various Tables.jl-provided tables work on really wide tables (dict tables, matrix, etc.).

bkamins:
Great work - thank you!

src/Tables.jl
```diff
@@ -402,44 +404,71 @@ are unknown (usually not inferrable). This is similar to the `Base.EltypeUnknown
...
+        return Schema{Tuple(map(sym, names)), Tuple{types...}}()
+    end
+end
 
-function Base.show(io::IO, sch::Schema{names, types}) where {names, types}
+function Base.show(io::IO, sch::Schema)
     get(io, :print_schema_header, true) && println(io, "Tables.Schema:")
-    Base.print_matrix(io, hcat(collect(names), types === nothing ? fill(nothing, length(names)) : collect(fieldtype(types, i) for i = 1:fieldcount(types))))
+    nms = sch.names
+    Base.print_matrix(io, hcat(nms isa Vector ? nms : collect(nms), sch.types === nothing ? fill(nothing, length(nms)) : collect(sch.types)))
```
bkamins, line 456: just to be sure - are both paths tested?

bkamins (approved):

PR #244, opened 2021-07-30
treat vectors as 1-column matrices for table(x)

### LINKED ISSUE #243: support AbstractVecOrMat in matrix.jl
If you have an `AbstractVector`, it would be convenient to be able to treat it as a table with a single column via `table(x)`.   It seems like this could be implemented simply by replacing `AbstractMatrix` with `AbstractVecOrMat` in `src/matrix.jl`.

(Recall that Julia vectors can be indexed like a matrix, with two indices, as long as the second index is `1`, and that `size(x,2)` returns `1` when `x` is a vector.  This is precisely so that vectors can be treated as 1-column matrices.)

stevengj:
Closes #243.

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -1,10 +1,10 @@
-istable(::AbstractMatrix) = false
-istable(::Type{<:AbstractMatrix}) = false
+istable(::AbstractVecOrMat) = false
+istable(::Type{<:AbstractVecOrMat}) = false
 
-rows(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))
-columns(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))
+rows(m::T) where {T <: AbstractVecOrMat} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractVecOrMat as a table"))
+columns(m::T) where {T <: AbstractVecOrMat} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractVecOrMat as a table"))
 
-struct MatrixTable{T <: AbstractMatrix} <: AbstractColumns
+struct MatrixTable{T <: AbstractVecOrMat} <: AbstractColumns
     names::Vector{Symbol}
     lookup::Dict{Symbol, Int}
     matrix::T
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -287,6 +287,14 @@ end
...
     # test for presence of superfluous wrapping
     @test typeof(X′) == typeof(X)
+
+    # Treating vectors as 1-column tables:
+    x = [1,2,3,4,5]
+    tbl = Tables.Columns(Tables.table(x))
+    @test keys(tbl) == [:Column1]
+    x′ = Tables.matrix(tbl)
+    @test x′ == reshape(x, :, 1)
+    @test Tables.matrix(Tables.table(x)) == x
```
bkamins: are you convinced `matrix ∘ table` should be no-op? I think people would assume that `matrix` would guarantee to return a matrix.

src/matrix.jl
```diff
@@ -1,10 +1,10 @@
-istable(::AbstractMatrix) = false
-istable(::Type{<:AbstractMatrix}) = false
+istable(::AbstractVecOrMat) = false
+istable(::Type{<:AbstractVecOrMat}) = false
 
-rows(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))
-columns(m::T) where {T <: AbstractMatrix} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractMatrix as a table"))
+rows(m::T) where {T <: AbstractVecOrMat} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractVecOrMat as a table"))
+columns(m::T) where {T <: AbstractVecOrMat} = throw(ArgumentError("a '$T' is not a table; see `?Tables.table` for ways to treat an AbstractVecOrMat as a table"))
```
bkamins: I think the changes for `rows` and `columns` are breaking and `istable` already returned `false` AFAICT. As commented earlier we currently allow:
```
julia> Tables.columns([(1,2),(3,4)])
Tables.CopiedColumns{NamedTuple{(Symbol("1"), Symbol("2")), Tuple{Vector{Int64}, Vector{Int64}}}} with 2 rows, 2 columns, and schema:
 Symbol("1")  Int64
 Symbol("2")  Int64
```

test/runtests.jl
```diff
@@ -287,6 +287,13 @@ end
     @test X′ == X
     # test for presence of superfluous wrapping
     @test typeof(X′) == typeof(X)
+
+    # Treating vectors as 1-column tables:
+    x = [1,2,3,4,5]
+    tbl = Tables.Columns(Tables.table(x))
```
bkamins, line 293: also `Tables.table` with the second positional argument would be good to test.

test/runtests.jl
stevengj: changed

src/matrix.jl
stevengj: I reverted the changes to these functions.

test/runtests.jl
stevengj: What is the second positional argument?  I only see one method of `table`, with only a single positional argument.  Do you mean the `header` keyword argument?

bkamins: Ah - right. I meant `header`.

Sorry for mishaps (@quinnj knows the Tables.jl best so he probably would not make the mistakes I make, but I know he currently needs some help in handling the requests so I thought that I comment to move things forward a bit before he can get to it).

stevengj: Added a `header` test.

quinnj (approved):
Looks fine by me! I like the solution of going through `Tables.matrix` since it's still an explicit operation.

Issue #174, opened 2020-05-21
Row(x::AbstractRow) = x


tkf:
I think this definition has some problems:

https://github.com/JuliaData/Tables.jl/blob/53dfc6c6d63f7f7aaee9ad4f4e52d447de057459/src/Tables.jl#L217

I think it's better to maintain the invariance `T(args..) isa T` if `T isa Type` [even though Julia documentation says it is OK to not do this "in _very rare_ cases"](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Constructors-that-don't-return-instances-of-their-own-type-1).

For example, due to this, `Tables.Row` interacts badly with `Base.Generator`:

```julia
julia> struct MyRow <: Tables.AbstractRow
           x
       end

julia> collect(Tables.Row(x) for x in [MyRow(1)]) |> typeof
Array{MyRow,1}

julia> collect(Tables.Row(x) for x in MyRow[]) |> typeof
Array{Tables.Row,1}
```

Notice that the element type of the last array is `Tables.Row` even though this generator cannot possibly produce a `Tables.Row`.

This is because [`Base.@default_eltype`](https://github.com/JuliaLang/julia/blob/9d70d450e771236b7e1e7cb0c7ac965a8843d662/base/array.jl#L647-L649) has a special case for `Base.Generator`.  It is better to fix this in `Base` but I think it is a good example that it is important to maintain `T(args..) isa T`.

I think it might be better to turn `Row` to an abstract type and soft-deprecate `AbstractRow`.  Something like

```julia
abstract type Row end
const AbstractRow = Row  # to be removed in 2.0

struct GenericRow{T} <: Row
    x::T
end

Row(x::Row) = x
Row(x) = GenericRow(x)
```

nalimilan:
There are other `Abstract*` types in the package, so `AbstractRow` is more explicit than `Row`. How about making `AbstractRow` a kind of constructor instead?

tkf:
I agree making `AbstractRow` callable is a valid option. I just thought it'd be better to make "consumer-facing" API (as opposed to implementer-facing) like `Row` to be a bit more terse.

There are only two more types with `Abstract` prefix:

* `AbstractColumns`: Similar soft-deprecation strategy is possible.
* `AbstractRowTable`: It's defined as `const AbstractRowTable = AbstractVector{T} where {T <: AbstractRow}` and it looks like just an internal type alias. So I don't think it counts as a "consistency penalty."

[...]

PR #255, opened 2021-09-22
Don't lie when calling Tables.Row/Tables.Columns

### LINKED ISSUE #254: `Columns` is a lie
Calling `Columns` on a table does not always return something that `isa Columns`:

```julia
julia> using Tables

julia> rowdata = [(y=i, x=2i) for i in 1:10]
10-element Vector{NamedTuple{(:y, :x), Tuple{Int64, Int64}}}:
 (y = 1, x = 2)
 (y = 2, x = 4)
 (y = 3, x = 6)
 (y = 4, x = 8)
 (y = 5, x = 10)
 (y = 6, x = 12)
 (y = 7, x = 14)
 (y = 8, x = 16)
 (y = 9, x = 18)
 (y = 10, x = 20)

julia> coldata = Tables.Columns(Tables.columns(rowdata))
Tables.CopiedColumns{NamedTuple{(:y, :x), Tuple{Vector{Int64}, Vector{Int64}}}} with 10 rows, 2 columns, and schema:
 :y  Int64
 :x  Int64

julia> coldata isa Tables.Columns
false

julia> coldata isa Tables.CopiedColumns
true
```

For context, I'd hoped to replace `ColumnTable` with `Columns` in StatsModels, since `ColumnTable` packs a lot of extra information into the type parameters, being a named tuple.  But we currently use `NamedTuple` / `ColumnTable` as a dispatch target for methods of e.g. `modelcols`, and define a generic `modelcols(term, data) = modelcols(term, Tables.columntable(data))`.  I thought first that I could use `getcolumn` on any table, but that doesn't work with tables that where `columnaccess` is false.  The next thought was that I could replace `columntable` with `Columns(columns(data))` and `::ColumnTable` with `::Columns`, and that's what led me here...

quinnj:
Fixes #254. The problem with returning the input itself for any
`::AbstractColumns` is that it makes it hard to use `Tables.Columns` as
a dispatchable type when desired. If someone passes an object that's
already `::AbstractColumns`, it's also not the end of the world to wrap
it in one more layer where `getcolumn` calls are probably going to get
inlined away anyway.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -259,7 +259,10 @@ to provide useful default behaviors (allows any `AbstractColumns` to be used lik
 struct Columns{T} <: AbstractColumns
     x::T
 
-    Columns(x::T) where {T} = new{T}(columns(x))
+    function Columns(x)
+        cols = columns(x)
+        return new{typeof(cols)}(cols)
+    end
 end
 
 Columns(x::Columns) = x
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -244,7 +244,7 @@ struct Row{T} <: AbstractRow
     x::T
 end
 
-Row(x::AbstractRow) = x
+Row(x::Row) = x
 
 """
     Tables.Columns(columns)
@@ -260,7 +260,7 @@ struct Columns{T} <: AbstractColumns
     x::T
 end
 
-Columns(x::AbstractColumns) = x
+Columns(x::Columns) = x
 
 # Columns can only wrap something that is a table, so we pass the schema through
 schema(x::Columns) = schema(getx(x))
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -258,6 +258,8 @@ to provide useful default behaviors (allows any `AbstractColumns` to be used lik
 """
 struct Columns{T} <: AbstractColumns
     x::T
+
+    Columns(x::T) where {T} = new{T}(columns(x))
 end
 
 Columns(x::Columns) = x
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Why can't users just dispatch on `AbstractColumns`?

quinnj:
When implementing the `Tables.columns(x)` interface, you're not _required_ to subtype `AbstractColumns`; for example, a `DataFrame` defines `Tables.columns(x::DataFrame) = x`, but it doesn't subtype `AbstractColumns`. But as you've noted, a developer could still dispatch on `AbstractColumns` as long as they first called `Tables.Columns(x)` first, which would be guaranteed to return an `AbstractColumns` object (either `Columns` itself, or some subtype). But as noted in #174, this has other issues since the return type become unpredictable (effectively `Union{Columns, AbstractColumns}`), as well as violating the usual Julia idiom that `T(x)` should return a `T`.

kleinschmidt:
This might be slightly out of scope for this PR, but would it also make sense to add a method like `Columns(x) = Columns(columns(x))`?  What will happen now if you pass something that's not `isa AbstractColumns` to `Columns`?

quinnj:
Yes, I do think we should make that change. The problem currently is people probably _assume_ it's calling `Tables.columns`, so they would get an equally confusing error by having a non-columns object wrapped in `Columns`. I'm remembering now that we also discussed (@nalimilan and @bkamins and I) having a `Tables.Rows` object that was a similar wrapper like `Tables.Columns` that called `Tables.rows` for you and was a dispatchable object.

quinnj:
Alright; going for it. Going to tag a release and make a couple of PRs to packages that can update their usage of `Tables.Columns` now.

PR #259, opened 2021-10-16
Minor docs clarification for custom types

### LINKED ISSUE #257: columnaccess documentation clarification
I was struggling a bit with implementing the interface for a column-table-like type, and eventually tracked it down to this part of the docs:

https://tables.juliadata.org/stable/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source)

> Tables.columnaccess(table) |   | Declare that your table type defines a Tables.columns(table) method

I think one actually needs to define `Tables.columnaccess(::Type{MyTableType})`. The docstring of `columnaccess` doesn't mention this either from my reading. There is a default implementation `columnaccess(::T) where {T} = columnaccess(T)` [here](https://github.com/JuliaData/Tables.jl/blob/main/src/Tables.jl#L334-L335).

The method accepting the type is called in a number of places, for example [in rows()](https://github.com/JuliaData/Tables.jl/blob/main/src/fallbacks.jl#L87), so since I'd only defined `columnaccess(::MyTableType)`, `rows` didn't try to use the `RowIterator`.

Have I understood correctly? If so I'm happy to draft a PR to clarify the docs.

tpgillam:
Closes https://github.com/JuliaData/Tables.jl/issues/257

This attempts to clarify those places where interface implementers are expected to declare methods taking their type, rather than instances of their type.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -225,20 +225,20 @@ Tables.Columns
 Now that we've seen how one _uses_ the Tables.jl interface, let's walk-through how to implement it; i.e. how can I
 make my custom type valid for Tables.jl consumers?
 
-The interface to becoming a proper table is straightforward:
+For a type `MyTable`, the interface to becoming a proper table is straightforward:
 
-| Required Methods             | Default Definition           | Brief Description                                                                                                               |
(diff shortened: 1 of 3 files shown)


quinnj (approved):
Thank you!

PR #262, opened 2021-11-18
Fix inconsistency between row and column behavior of MatrixTable

### LINKED ISSUE #261: BoundsError on `last(::MatrixTable)`
```
julia> last(Tables.table(rand(3,2)))
ERROR: BoundsError: attempt to access 3×2 Matrix{Float64} at index [1:3, 3]
Stacktrace:
 [1] throw_boundserror(A::Matrix{Float64}, I::Tuple{Base.Slice{Base.OneTo{Int64}}, Int64})
   @ Base .\abstractarray.jl:651
 [2] checkbounds
   @ .\abstractarray.jl:616 [inlined]
 [3] view
   @ .\subarray.jl:177 [inlined]
 [4] getcolumn
   @ C:\Users\sternlab\.julia\packages\Tables\OWzlh\src\matrix.jl:39 [inlined]
 [5] getindex
   @ C:\Users\sternlab\.julia\packages\Tables\OWzlh\src\Tables.jl:179 [inlined]
 [6] last(a::Tables.MatrixTable{Matrix{Float64}})
   @ Base .\abstractarray.jl:437
 [7] top-level scope
   @ REPL[10]:1
```
This breaks display of `MatrixTable` in Pluto.

quinnj:
Fixes #261. Because `MatrixTable` subtypes `AbstractColumns`, it
inherits certain behavior like `length`; but when it needs to fulfill
the "rows" interface, it needs `length` to be defined in terms of the #
of rows instead of the # of columns. We fix this inconsistency here by
defining another type that mirrors `MatrixTable` but is used in the rows
case.

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -10,13 +10,24 @@ struct MatrixTable{T <: AbstractVecOrMat} <: AbstractColumns
     matrix::T
 end
 
-isrowtable(::Type{<:MatrixTable}) = true
-names(m::MatrixTable) = getfield(m, :names)
+struct MatrixRowTable{T <: AbstractVecOrMat}
+    names::Vector{Symbol}
+    lookup::Dict{Symbol, Int}
+    matrix::T
+end
+
+const MatrixTables{T} = Union{MatrixTable{T}, MatrixRowTable{T}}
+
+names(m::MatrixTables) = getfield(m, :names)
 
 # row interface
+istable(::Type{<:MatrixTable}) = true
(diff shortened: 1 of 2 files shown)


bkamins:
I understand one needs to call `rows` on `MatrixTable` to get a row iterator - right?

quinnj:
Correct; that was always the _official_ API, but if you happened to just iterate `MatrixTable`, you would get rows. That's the change in behavior with this PR.

bkamins (approved):

PR #263, opened 2021-11-27
Avoid assumption that `Tables.getcolumn(x)` always returns an AbstractVector.

OkonSamuel:
From the documentation, calling `Tables.getcolumn` on an `AbstractColumns` object should return an `AbstractColumn` object (i.e any object `col` that supports `length(col)` and `col[i]` for any `i = 1:length(col)`). But these objects aren't necessarily instances of `AbstractVectors`. In the case that `Tables.getcolumn` doesn't return an instance of `AbstractVector` the method `Tables.matrix` fails and throws the following error.
```julia
ERROR: ArgumentError: indexed assignment with a single value to many locations is not supported; perhaps use broadcasting `.=` instead?
```

Below is a sample code that leads to the above error.
```julia
struct MockTable end
Tables.istable(::Type{MockTable}) = true
Tables.columnaccess(::Type{MockTable})) = true
Tables.columnnames(x::MockTable) = (:a, :b, :c)
Tables.columns(x::MockTable) = x
Tables.getcolumn(x::MockTable, ::Symbol) = (1, 2, 3)
Tables.getcolumn(x::MockTable, ::Int) = (1, 2, 3)
Tables.schema(x::MockTable) = Tables.Schema((:a, :b, :c), NTuple{3, Int})
Tables.matrix(MockTable())
```
The error is non-existent after this PR.

### DIFF
--- a/src/matrix.jl
+++ b/src/matrix.jl
@@ -77,12 +77,12 @@ function matrix(table; transpose::Bool=false)
     if !transpose
         matrix = Matrix{T}(undef, n, p)
         for (i, col) in enumerate(cols)
-            matrix[:, i] = col
+            matrix[:, i] .= col
         end
     else
         matrix = Matrix{T}(undef, p, n)
         for (i, col) in enumerate(cols)
-            matrix[i, :] = col
+            matrix[i, :] .= col
         end
     end
     return matrix
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -206,6 +206,16 @@ end
     @test Tables.materializer(1) === Tables.columntable
 end
 
+# Table with tuple columns.
+struct MockTable end 
+Tables.istable(::Type{MockTable}) = true
+Tables.columnaccess(::Type{MockTable}) = true
+Tables.columnnames(x::MockTable) = (:a, :b, :c)
+Tables.columns(x::MockTable) = x
+Tables.getcolumn(x::MockTable, ::Symbol) = (1, 2, 3)
+Tables.getcolumn(x::MockTable, ::Int) = (1, 2, 3)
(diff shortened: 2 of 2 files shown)


quinnj:
Does anyone understand how the "under the hood" implementation is different here? i.e. setting a column in a matrix via `m[i, :] = col` vs. `m[i, :] .= col`? Obviously the latter does broadcasting, and I assume is element-wise, but perhaps `m[i, :] = col` is able to do the equivalent of a `copyto!`/`memcpy` for better performance? @bkamins or @nalimilan might know better.

Anyway, I think this change should be fine, but it'd nice to do a little benchmarking to see if there's a significant change in performance and to add a test to this PR. Even just including the code in the original post here seems like it would suffice.

Thanks for the contribution @OkonSamuel!

nalimilan:
This indeed looks more correct. I haven't checked, but I think `m[i, :] .= col` will hit this optimized `copyto!` method (via `materialize!`): https://github.com/JuliaLang/julia/blob/master/base/broadcast.jl#L948-L956
```julia
julia> Meta.@lower m[1, :] .= [1]
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope`
1 ─ %1 = Base.dotview(m, 1, :)
│   %2 = Base.vect(1)
│   %3 = Base.broadcasted(Base.identity, %2)
│   %4 = Base.materialize!(%1, %3)
└──      return %4
))))
```

nalimilan:
Ah and it would be good to add a test, otherwise we may break this without noticing in the future.

OkonSamuel:
@nalimilan, @quinnj I just added tests as required. Sorry I have been quite busy.

nalimilan (approved):

quinnj (approved):
Thanks @OkonSamuel!

PR #266, opened 2022-01-02
Fix typo in docs excerpt from DataFrames.jl source code

Jasha10:
The Tables.jl docs make reference to the DataFrames.jl source code.

The quoted source code has some errors in it (and the quote disagrees with the actual source). This PR fixes those errors.


bkamins:
The code still disagrees with https://github.com/JuliaData/DataFrames.jl/blob/72cb858d6f92887ed8d2d75f94108e2a864d2a5c/src/other/tables.jl#L49 part, but maybe this can be skipped (this part of the code is made to ensure we do not use Tables.jl mechanics when it is not desirable to use it).

Issue #274, opened 2022-02-20
Make Tables.dictrowtable use OrderedDict


bkamins:
It is debatable, but maybe it is better for `Tables.dictrowtable` to preserve the order of appearance of columns in the input by switching it to use `OrderedDict` internally?

quinnj:
I agree this would be more consistent now that we use `OrderedDict` for `dictcolumntable`; perhaps we can request @mathieu17g to make a PR?

bkamins:
Ah - indeed.

@mathieu17g - in #275 two things are missed:
* you updated `dictcolumntable` but also `dictrowtable` should be likewise updated
* the docstring of `dictcolumntable` should be fixed after your PR (as it mentions `Dict` incorrectly now)

would you be willing to make this PR?

mathieu17g:
Yes I will look into it.

Issue #273, opened 2022-02-11
Performance issue with `Tables.columns` fallback functions at 1st run with `Tables.Schema = nothing` and `Base.IteratorSize = SizeUnknown()`


mathieu17g:
Hello,

I'm not aware if any user implementing the Tables interface and relying on fallback functions with no `Schema` has already encountered a performance issue at compilation, but here is below a particular case encountered in ArchGDAL.jl

Using the Tables.jl interface on ArchGDAL.jl layer objects, @visr stumbled on a performance issue with `Tables.columns` fallback functions at 1st run,  with `Tables.Schema = nothing` and `Base.IteratorSize = SizeUnknown()`. The performance bottleneck at first run is located in `Tables.add_or_widen!` (seen when profiling). Probably around the columm allocation machinery based on `Tuple`.

There is no problem when providing detailed types in the `Schema`. Unfortunately, we cannot know the tightest schema for sure from a layer feature definition.

The problem can be reproduced with sequence below: **~170s at first run** for a 10 row table with more than 300 columns

Download [BasinATLAS_v10_lev01.zip](https://github.com/JuliaData/Tables.jl/files/8046182/BasinATLAS_v10_lev01.zip) and decompress it in `./BasinATLAS_v10_lev01/`

```julia
julia> import ArchGDAL as AG

julia> using DataFrames

julia> using Profile

julia> @time @profile DataFrame(AG.getlayer(AG.read("BasinATLAS_v10_lev01")))
172.879812 seconds (53.38 M allocations: 2.992 GiB, 0.44% gc time, 99.38% compilation time)
10×295 DataFrame
 Row │                            HYBAS_ID    NEXT_DOWN  NEXT_SINK   MAIN_BAS    DIST_SINK  DIST_MAIN  SUB_AREA   UP_AREA    PFAF_ID  EN ⋯
     │ IGeometr…                  Int64       Int64      Int64       Int64       Float64    Float64    Float64    Float64    Int64    In ⋯
─────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ Geometry: wkbMultiPolygon  1010000010          0  1010000010  1010000010        0.0        0.0  2.99531e7  2.99531e7        1     ⋯
   2 │ Geometry: wkbMultiPolygon  2010000010          0  2010000010  2010000010        0.0        0.0  1.78589e7  1.78589e7        2
[...]

PR #275, opened 2022-02-23
Inline of `replacex` in `add_or_widen!` (compilation performance) and use of `OrderedDict` in `dictcolumntable` (column ordering)

mathieu17g:
In case the propositions made in https://github.com/JuliaData/Tables.jl/issues/273#issuecomment-1037986769 and https://github.com/JuliaData/Tables.jl/issues/274#issue-1145052654 are accepted, this PR can solve issues #273 and #274


bkamins:
Looks good, but @quinnj knows the details.

bkamins:
@quinnj - could you have a look at this PR please?

quinnj (approved):
Hmmmm, yes, I guess this is all fine. The inline of `replacex` is obviously fine; for the `OrderedDict` change, I would note that we've typically directed users to only rely on the column ordering provided by the _schema_ instead of any "default" table order. Though in the dict case I can see how using an OrderedDict would be quite useful/convenient and so I'm in favor.

It does make me wonder if we should clarify in the docs about this: should table formats be _required_ to store columns in a specific order? We don't currently because the interface already exposes `getcolumn(cols, i)`, so that _feels_ like you're getting the columns in order.

bkamins:
Given we say that:

> The methods taking a single `Symbol` or `Int` are both required for the `AbstractColumns` and `AbstractRow` interfaces

I think we should require that the order is somehow guaranteed. In other words - since we allow getting column by integer it would be good to clarify what this means in general (not only by default - which is defined to use `getfield`).

PR #279, opened 2022-04-06
Use columntable in datavaluerows for col sources

davidanthoff:
@quinnj, the idea here would be that https://github.com/JuliaData/DataFrames.jl/blob/b1fdb7621fa38c381dad928317ec1de562d09e71/src/other/tables.jl#L93 could be rewritten as
```julia
IteratorInterfaceExtensions.getiterator(df::AbstractDataFrame) = Tables.datavaluerows(df)
```
But I'm not entirely sure this is right :) I think this PR only makes sense if the idea is that for any table that is primarily column based, one should first create a columntable before one calls the `datavaluerows` function. Is that so? I don't really understand the underlying mechanics there.

The benefit of this PR would be that now the instructions for any Tables.jl source that wants to also enable TableTraits.jl integration would be a bit simpler, i.e. the instructions would be to just always add
```julia
IteratorInterfaceExtensions.getiterator(x::MyType) = Tables.datavaluerows(x)
IteratorInterfaceExtensions.isiterable(::MyType) = true
TableTraits.isiterabletable(::MyType) = true
```
and that would work regardless of whether the source internally stores things as columns or rows.

### DIFF
--- a/src/tofromdatavalues.jl
+++ b/src/tofromdatavalues.jl
@@ -93,7 +93,15 @@ IteratorInterfaceExtensions.getiterator(x::MyTable) = Tables.datavaluerows(x)
 ```
 """
 function datavaluerows(x)
-    r = Tables.rows(x)
+    x2 = if Tables.columnaccess(x)
+        Tables.columntable(x)
+    elseif Tables.rowaccess(x)
+        x
+    else
+        error("Invalid input table, must either declare to have a Tables.rows or Tables.columns method.")
+    end
+
+    r = Tables.rows(x2)
     s = Tables.schema(r)
     s === nothing && error("Schemaless sources cannot be passed to datavaluerows.")
     return DataValueRowIterator{datavaluenamedtuple(s), typeof(s), typeof(r)}(r)
--- a/src/tofromdatavalues.jl
+++ b/src/tofromdatavalues.jl
@@ -93,15 +93,7 @@ IteratorInterfaceExtensions.getiterator(x::MyTable) = Tables.datavaluerows(x)
 ```
 """
 function datavaluerows(x)
-    x2 = if Tables.columnaccess(x)
-        Tables.columntable(x)
-    elseif Tables.rowaccess(x)
-        x
-    else
-        error("Invalid input table, must either declare to have a Tables.rows or Tables.columns method.")
-    end
-
-    r = Tables.rows(x2)
+    r = Tables.rows(Tables.columnaccess(x) ? Tables.columntable(x) : x)
     s = Tables.schema(r)
     s === nothing && error("Schemaless sources cannot be passed to datavaluerows.")
     return DataValueRowIterator{datavaluenamedtuple(s), typeof(s), typeof(r)}(r)


davidanthoff:
Actually, just out of curiosity: why is the call to `Tables.columntable` needed in https://github.com/JuliaData/DataFrames.jl/blob/b1fdb7621fa38c381dad928317ec1de562d09e71/src/other/tables.jl#L94 in the currently shipping version?

bkamins:
> why is the call to `Tables.columntable` needed

I have not implemented it, but it turns type-unstable to type-stable container.

quinnj:
> Actually, just out of curiosity: why is the call to `Tables.columntable` needed in https://github.com/JuliaData/DataFrames.jl/blob/b1fdb7621fa38c381dad928317ec1de562d09e71/src/other/tables.jl#L94 in the currently shipping version?

This is necessary for performance in Query.jl, where QueryOperators expect type-stable rows. At one point, we had the case where we were calling `NamedTuple(::DataFrameRow)` and it was too expensive and hurt performance in Query framework. We then had the discussion that Query.jl is really only geared towards narrow tables that can return something fully type-stable from `getiterator` (i.e. concrete `eltype`), so hence the change to call `Tables.columntable`.

src/tofromdatavalues.jl
```diff
@@ -93,7 +93,15 @@ IteratorInterfaceExtensions.getiterator(x::MyTable) = Tables.datavaluerows(x)
...
-    r = Tables.rows(x)
+    x2 = if Tables.columnaccess(x)
+        Tables.columntable(x)
+    elseif Tables.rowaccess(x)
+        x
+    else
+        error("Invalid input table, must either declare to have a Tables.rows or Tables.columns method.")
+    end
+
+    r = Tables.rows(x2)
```
quinnj:
```suggestion
    r = Tables.rows(Tables.columnaccess(x) ? Tables.columntable(x) : x)
```
Hmmmm.....yeah, I think this is probably fine to change; and it shouldn't affect DataFrames because calling `columntable(::ColumnTable)` is a no-op.

The reason for not throwing an error in my suggestion is because someone may have, say, a `Vector{Any}` where elements are NamedTuples with `missing` values, but they want to use that in Query.jl. A `Vector{Any}` doesn't have `rowaccess` defined, but it will work with `Tables.rows`, assuming it actually has valid "rows" as elements.

davidanthoff: Yep, that makes sense!

davidanthoff:
@quinnj one more question: in the case where `Tables.columnaccess(x)==false`, this PR now would essentially amount to
```julia
    r = Tables.rows(x)
    s = Tables.schema(r)
    s === nothing && error("Schemaless sources cannot be passed to datavaluerows.")
    return DataValueRowIterator{datavaluenamedtuple(s), typeof(s), typeof(r)}(r)
```
but rereading your comment above, that would probably be type-instable, right?

The scenario that I have in the back of my mind is row wise reading from CSV.jl in combination with Query.jl. It would be nice if something like
```julia
CSV.Rows(filename) |> @filter(...) |> @select(...) |> DataFrame
```
would just work out of the box. My thinking had been that if we merge this PR here and then add:
```julia
IteratorInterfaceExtensions.getiterator(x::CSV.Rows) = Tables.datavaluerows(x)
IteratorInterfaceExtensions.isiterable(::CSV.Rows) = true
TableTraits.isiterabletable(::CSV.Rows) = true
```
it would all work, but now I'm thinking that we might end up with a type instable situation, right?

Is there some equivalent of `Tables.columntable` that we could use in the row iterating case that _doesn't_ materialize all the columns, but just provides an automatic type stable named tuple iterator around a row wise source like `CSV.Rows`?

quinnj:
> Is there some equivalent of `Tables.columntable` that we could use in the row iterating case that _doesn't_ materialize all the columns, but just provides an automatic type stable named tuple iterator around a row wise source like `CSV.Rows`?

I think for the rows case, we just want to wrap them as-is with `IteratorInterfaceExtensions.getiterator(x::CSV.Rows) = Tables.datavaluerows(x)` as you specified above; the point is that `DataValueRowIterator` _is_ type-stable and turns any Tables.jl-compatible "row" object into a fully-typed NamedTuple w/ DataValue for `missing`. So I think this is PR is fine as-is. We could probably do some performance testing to confirm, but I believe that would be essentially the same perf testing I did originally when adding this functionality to Tables.jl to cover the Query.jl cases for convenience. Really the main value of this PR is allow column-based sources to more automatically be type-stable.

quinnj:
Closing as old/stale

PR #284, opened 2022-06-28
add getrows

CarloLucibello:
as per comment https://github.com/JuliaData/Tables.jl/pull/278#issuecomment-1132706343

@quinnj @bkamins @nalimilan @ablaom @ToucheSir

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -570,9 +570,11 @@ end
 
 Return one or more rows from table `x` according to the position(s) specified by `inds`:
 
-- If `inds` is a single integer return a row object.
-- If `inds` is a collection of integers, return an indexable object of rows. 
+- If `inds` is a single non-boolean integer return a row object.
+- If `inds` is a vector of non-boolean integers, a vector of booleans, or a `:`, return an indexable object of rows. 
   In this case, the returned type is not necessarily the same as the original table type.
+  
+If other type of `inds` is passed than specified above the behavior is undefined.
 
 The `view` argument influences whether the returned object is a view of the original table
 or an independent copy:
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -566,12 +566,12 @@ struct Partitioner{T}
 end
 
 """
-    getrows(x, inds; view=nothing)
+    Tables.getrows(x, inds; view=nothing)
 
 Return one or more rows from table `x` according to the position(s) specified by `inds`:
 
 - If `inds` is a single integer return a row object.
-- If `inds` is a collection of integers, return a table object. 
+- If `inds` is a collection of integers, return an indexable object of rows. 
   In this case, the returned type is not necessarily the same as the original table type.
 
 The `view` argument influences whether the returned object is a view of the original table
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -583,7 +583,7 @@ or an independent copy:
   This applies both to returning a row or a table.
  
 Any specialized implementation of `getrows` must support the `view=nothing` argument. 
-Support for `view=true` or `view=false` instead can be an opt-in 
+Support for `view=true` or `view=false` is optional
 (i.e. implementations might error on them if they are not supported).
 """
 function getrows end
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -566,6 +566,28 @@ struct Partitioner{T}
 end
 
 """
+    getrows(x, inds; view=nothing)
+
+Return one or more rows from table `x` according to the position(s) specified by `inds`.
+
+- If `inds` is a single integer return a row object.
+- If `inds` is a collection of integers, return a table object. 
+  In this case,t he returned type is not necessarily the same as the original table type.
+
+The `view` argument influences whether the returned object is a view of the original table
+or an independent copy:
+- If `view=nothing` (the default) then the implementation for a specific table type 
+  is free to decide  whether to return a copy or a view.
+- If `view=true` then a view is returned and if `view=false` a copy is returned.
+  This applies both to returning a row or a table.
+ 
+Any specialized implementation of `getrows` must support the `view=nothing` argument. 
+Support for `view=true` or `view=false` instead can be an opt-in 
+(i.e. implementations might error on them if they are not supported).
+"""
+function getrows end
+
+"""
     Tables.partitioner(f, itr)
     Tables.partitioner(x)
 
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -568,14 +568,15 @@ end
 """
     getrows(x, inds; view=nothing)
 
-Return one or more rows from table `x` according to the position(s) specified by `inds`.
+Return one or more rows from table `x` according to the position(s) specified by `inds`:
 
 - If `inds` is a single integer return a row object.
 - If `inds` is a collection of integers, return a table object. 
-  In this case,t he returned type is not necessarily the same as the original table type.
+  In this case, the returned type is not necessarily the same as the original table type.
 
 The `view` argument influences whether the returned object is a view of the original table
 or an independent copy:
+
 - If `view=nothing` (the default) then the implementation for a specific table type 
   is free to decide  whether to return a copy or a view.
 - If `view=true` then a view is returned and if `view=false` a copy is returned.
--- a/src/namedtuples.jl
+++ b/src/namedtuples.jl
@@ -108,7 +108,7 @@ end
 
 function getrows(x::RowTable, inds; view::Union{Bool,Nothing} = nothing)
     if view === true
-        return view(x, inds)
+        return Base.view(x, inds)
     else
         return x[inds]
     end
@@ -184,7 +184,7 @@ columntable(x::ColumnTable) = x
 
 function getrows(x::ColumnTable, inds; view::Union{Bool,Nothing} = nothing)
     if view === true
-        return map(c -> view(c, inds), x)
+        return map(c -> Base.view(c, inds), x)
     else
         return map(c -> c[inds], x)
     end
--- a/src/namedtuples.jl
+++ b/src/namedtuples.jl
@@ -106,6 +106,14 @@ function rowtable(itr::T) where {T}
     return collect(namedtupleiterator(eltype(r), r))
 end
 
+function getrows(x::RowTable, inds; view::Union{Bool,Nothing} = nothing)
+    if view === true
+        return view(x, inds)
+    else
+        return x[inds]
+    end
+end
+
 # NamedTuple of arrays of matching dimensionality
 const ColumnTable = NamedTuple{names, T} where {names, T <: NTuple{N, AbstractArray{S, D} where S}} where {N, D}
 rowcount(c::ColumnTable) = length(c) == 0 ? 0 : length(c[1])
@@ -173,3 +181,11 @@ function columntable(itr::T) where {T}
     return columntable(schema(cols), cols)
 end
 columntable(x::ColumnTable) = x
+
+function getrows(x::ColumnTable, inds; view::Union{Bool,Nothing} = nothing)
+    if view === true
+        return map(c -> view(c, inds), x)
+    else
+        return map(c -> c[inds], x)
+    end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 Manifest.toml
+.vscode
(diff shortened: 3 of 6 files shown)


src/Tables.jl
```diff
@@ -565,6 +565,28 @@ struct Partitioner{T}
...
 end
 
+"""
+    getrows(x, inds; view=nothing)
+
+Return one or more rows from table `x` according to the position(s) specified by `inds`.
+
+- If `inds` is a single integer return a row object.
+- If `inds` is a collection of integers, return a table object. 
+  In this case,t he returned type is not necessarily the same as the original table type.
```
bkamins:
```suggestion
  In this case, the returned type is not necessarily the same as the original table type.
```

bkamins:
This PR should probably provide appropriate implementations for `rowtable`, `columntable`, `dictrowtable`, `dictcolumntable`, `table` as these are defined in Tables.jl (an additional benefit of doing this will be that package developers will see how to properly implement `getrows` for their own types).

Also probably it would be good to bump version number so that it can be released after merging.

CarloLucibello:
> This PR should probably provide appropriate implementations for `rowtable`, `columntable`, `dictrowtable`, `dictcolumntable`, `table` as these are defined in Tables.jl (an additional benefit of doing this will be that package developers will see how to properly implement `getrows` for their own types).

I'm totally unfamiliar with this package so I was hoping to just get the stub in to get things started. I can have a look around and see if I can manage to write those implementations as well.

bkamins:
> I can have a look around and see if I can manage to write those implementations as well.

These should not be that hard. Of course we can do it later.

@quinnj, @nalimilan - do you think we can have some default fallback implementation of this method?

src/Tables.jl
```diff
@@ -565,6 +565,29 @@ struct Partitioner{T}
     x::T
 end
 
+"""
+    getrows(x, inds; view=nothing)
```
quinnj:
```suggestion
    Tables.getrows(x, inds; view=nothing)
```
We include the `Tables` prefix since this won't be exported.

src/Tables.jl
```diff
@@ -565,6 +565,29 @@ struct Partitioner{T}
...
+
+Return one or more rows from table `x` according to the position(s) specified by `inds`:
+
+- If `inds` is a single integer return a row object.
+- If `inds` is a collection of integers, return a table object.
```
quinnj:
```suggestion
- If `inds` is a collection of integers, return an indexable object of rows
```
Just a bit more specific; one of the main motivations for this was to ensure users can get an in-memory/indexable collection of rows in a consistent way.

src/namedtuples.jl
```diff
@@ -106,6 +106,14 @@ function rowtable(itr::T) where {T}
     return collect(namedtupleiterator(eltype(r), r))
 end
 
+function getrows(x::RowTable, inds; view::Union{Bool,Nothing} = nothing)
+    if view === true
+        return view(x, inds)
```
quinnj:
```suggestion
        return Base.view(x, inds)
```
Because we have `view` as the keyword argument, we need to qualify our reference to the _function_. It'd be great to add some tests for this.

src/namedtuples.jl
```diff
@@ -173,3 +181,11 @@ function columntable(itr::T) where {T}
     return columntable(schema(cols), cols)
 end
 columntable(x::ColumnTable) = x
+
+function getrows(x::ColumnTable, inds; view::Union{Bool,Nothing} = nothing)
+    if view === true
+        return map(c -> view(c, inds), x)
```
quinnj:
```suggestion
        return map(c -> Base.view(c, inds), x)
```
Same here

quinnj (changes requested):
This is a great start; thanks @CarloLucibello. Sorry I've been so slow on getting around to this or responding on the original issue. Just a little cleanup here and if you'll add some tests, I think it's ready to go. Once we merge, I can take a look at doing the implementation for the other table types in Tables.jl.

CarloLucibello:
Done! Can you unlock CI?

CarloLucibello:
good to go?

src/Tables.jl
```diff
@@ -565,6 +565,29 @@ struct Partitioner{T}
...
+The `view` argument influences whether the returned object is a view of the original table
+or an independent copy:
+
+- If `view=nothing` (the default) then the implementation for a specific table type 
+  is free to decide  whether to return a copy or a view.
+- If `view=true` then a view is returned and if `view=false` a copy is returned.
+  This applies both to returning a row or a table.
+ 
+Any specialized implementation of `getrows` must support the `view=nothing` argument. 
+Support for `view=true` or `view=false` instead can be an opt-in
```
```suggestion
Support for `view=true` or `view=false` is optional
```

src/Tables.jl
nalimilan: Are you sure we want to require `Tables.getrows` to return a indexable collection of rows? That would mean that e.g. for a `NamedTuple` of vectors `getrows` couldn't return another `NamedTuple` of vectors, even if that's the most efficient representation. IOW this is against the implementation that this PR adds for `ColumnTable`. :-)

nalimilan:
> @quinnj, @nalimilan - do you think we can have some default fallback implementation of this method?

Indeed, why not have `getrows` return a `RowTable` for row-oriented tables and a `ColumnTable` for column-oriented tables? For row-oriented tables, a simple solution would be to call `Tables.rows`, ensure the result is a vector (and if not call `collect` on it), and then call `getindex` or `view` on it. For column-oriented tables, call `Tables.columns`, and for each column ensure it is a vector (calling `collect` on it if necessary) and call `getindex` or `view` on it. In both cases, instead of calling `collect` it would be better in term of memory use to allocate a vector of the final size and fill it on the fly, but it's more tricky to do so maybe not worth it for a first fallback implementation.

EDIT: actually what I described is just calling `getrows(Tables.rowtable(tbl), ...))` when `Table.isrowtable(tbl)` and `getrows(Tables.columntable(tbl), ...))` otherwise. So that's quite trivial to implement.

ablaom:
@quinnj 🙏🏾

quinnj (approved):
This LGTM; @bkamins, can I ask you to take one more look here before merging? Sorry to let this sit for so long; I've been tied up with a lot f stuff.

src/Tables.jl
```diff
@@ -565,6 +565,29 @@ struct Partitioner{T}
...
 end
 
+"""
+    Tables.getrows(x, inds; view=nothing)
+
+Return one or more rows from table `x` according to the position(s) specified by `inds`:
+
+- If `inds` is a single integer return a row object.
+- If `inds` is a collection of integers, return an indexable object of rows. 
+  In this case, the returned type is not necessarily the same as the original table type.
```
bkamins:
```suggestion
- If `inds` is a single non-boolean integer return a row object.
- If `inds` is a vector of non-boolean integers, a vector of booleans, or a `:`, return an indexable object of rows. 
  In this case, the returned type is not necessarily the same as the original table type.
  
If other type of `inds` is passed than specified above the behavior is undefined.
```

The point is that under current definition:
```
julia> nt = (a=1, b=2, c=3)
(a = 1, b = 2, c = 3)

julia> rt = [nt, nt, nt]
3-element Vector{NamedTuple{(:a, :b, :c), Tuple{Int64, Int64, Int64}}}:
 (a = 1, b = 2, c = 3)
 (a = 1, b = 2, c = 3)
 (a = 1, b = 2, c = 3)

julia> getrows(rt, [1 2; 3 1])
2×2 Matrix{NamedTuple{(:a, :b, :c), Tuple{Int64, Int64, Int64}}}:
 (a = 1, b = 2, c = 3)  (a = 1, b = 2, c = 3)
 (a = 1, b = 2, c = 3)  (a = 1, b = 2, c = 3)
```

Alternatively to stating that other `inds` are undefined we could change the implementations to disallow them.

bkamins:
I have left one comment about allowed `inds` + Tables.jl version needs to be bumped.

src/Tables.jl
nalimilan: Good catch. It would be good to restrict the type signature of methods defined below to disallow undocumented index types. Otherwise, even if we say behavior is undefined some people may start relying on them.

bkamins: either restrict signatures or make a run-time check in the body of the function (I have not checked with would be better - the issue is that restricting signatures is very difficult for indexing). It might be easier to check if the output has a proper shape.

nalimilan:
@quinnj What about throwing an error as discussed [above](https://github.com/JuliaData/Tables.jl/pull/284#discussion_r956738499)?

bkamins:
I agree with @nalimilan that it is better to throw an error - at least for the reference implementations (so users, by looking at the code can see what is expected).

As I have commented above - I think the simplest thing to check is if the returned object is `RowTable` or `ColumnTable` respectively.

quinnj:
Ok, but here's an issue I've found digging into this further (and introducing the error checks mentioned); the current implementation of `getrows` for `ColumnTable` doesn't follow the `getrows` docs, since it returns a `ColumnTable` instead of `return an indexable object of rows.`.

The point of this is we always get either a single row or collection of rows, right? Even if the input is column-oriented? Or should we update the docs and say if `inds` is a collection, you get a subset of original input, row or column oriented.

quinnj:
And indeed, @nalimilan pointed this exact issue out [here](https://github.com/JuliaData/Tables.jl/pull/284#discussion_r912234537).

One idea is that we could provide a generic `ColumnsRows` struct that would basically be a wrapper for any result of `Tables.columns` that provided an "indexable collection of rows" interface to the underlying columns (essentially iterating `ColumnsRow` that we already use in the generic `Tables.rows` fallback for column oriented tables).

I might take a stab at this approach and see what it looks like.

bkamins:
> Or should we update the docs and say if `inds` is a collection, you get a subset of original input, row or column oriented.

This is what I think we should do. For example for DataFrames.jl I imagine that we will have the following implementation:
```
Tables.getrows(df::AbstractDataFrame, inds; view=nothing) = view === true ? view(df, inds, :) : df[inds, :]
```
as it seems most natural (and in general - it will be more efficient for a given table type to decide what to do; I assume that people will expect `Tables.getrows` to be efficient).

---

I would set the issue of "indexable collection of rows" and "indexable collection of columns" (for completeness as separate case) since it is orthogonal to the `Tables.getrows` design. I think we should provide such an API (again, in DataFrames.jl this API is provided by the `eachrow` and `eachcol` functions, but we could choose some other names here, as in Base Julia `eachrow` and `eachcol` were added later and they only support iteration interface but not indexing interface).

quinnj:
So given that we want this to be more of a general "subset" function, I'd like to change the name from `getrows`. 1) I think `Tables.getrows` is too similar to `Tables.rows` and it seems hard to explain the difference since they seem so similar and 2) for column-oriented tables, you're not really getting "rows" back, but you're just getting a subset of the original table.

I'm going to propose we call this `Tables.subset` to be more clear.

bkamins:
`Tables.subset` is very good for me as it is consistent in terminology with DataFrames.jl :).
@ablaom @ToucheSir @CarloLucibello - are you OK with this?

Alternatively we could have 2 functions:
* `Table.getrow` to get a single row (and it would accept a single index)
* `Table.subset` to get a subset of a table and return a table (and it would expect a collection of indices)

quinnj:
Ok, cleaned up implementation here: https://github.com/JuliaData/Tables.jl/pull/292

CarloLucibello:
`Tables.subset` is a reasonable choice, fine for me

ablaom:
Mmm. I'm not super happy with removing "rows" from the the name. If I'm new to tables in julia, "subset" could mean anything. I similarly never liked name like "select" . Select what? I want to know which axis I'm slicing on. 

How about `rowsubset` ?

quinnj:
But that's kind of the point of our discussion/decision above: this function isn't necessarily returning "rows" if the original table input is column-oriented; rather, it returns a "subset" of the original table, preserving whether the original table was column or row-oriented. Thus, `Tables.subset` (note the `Tables.` prefix!) seems clearest to what the function is actually doing.

ablaom:
Yes, I get that it is not returning "rows" in the sense of not returning a row iterator. But I'm not sure general users equate "rows" with "row iterator", maybe just Tables.jl developers do that 😉 . 

I also appreciate these name decisions are very difficult to get right and to get consensus on. Just putting in my two cents.

And I agree that aligning with DataFrames - which most users will know well - makes sense.

ToucheSir:
One could always go maximally explicit with something like `Tables.subset_along_row_dim`, but I imagine folks would not be terribly enthused about that name :P

PR #285, opened 2022-06-29
Add `nrow` and `ncol`

rikhuijzer:
I noticed that there is a `rowcount`, but not a `columncount` and that `length(columnnames(cols))` was used in the codebase to count the number of columns (because that is the fastest?). This PR suggests to add a `columncount` function.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -32,7 +32,7 @@ Interface definition:
 | `Tables.columnnames(table)`                              | propertynames(table)        | Return column names for a table as an indexable collection                                                                                                   |
 | **Optional methods**                                     |                             |                                                                                                                                                              |
 | `Tables.getcolumn(table, ::Type{T}, i::Int, nm::Symbol)` | Tables.getcolumn(table, nm) | Given a column eltype `T`, index `i`, and column name `nm`, retrieve the column. Provides a type-stable or even constant-prop-able mechanism for efficiency. |
-| `Tables.ncol(table)`                                    | Tables.ncol(table)          | Return the number of columns                                                                                                                                 |
+| `Tables.ncol(table)`                                     | Tables.ncol(table)          | Return the number of columns                                                                                                                                 |
 
 Note that subtypes of `Tables.AbstractColumns` **must** overload all required methods listed
 above instead of relying on these methods' default definitions.
@@ -66,7 +66,7 @@ Interface definition:
 | `Tables.columnnames(row)`                              | propertynames(row)        | Return column names for a row as an indexable collection                                                                                                         |
 | **Optional methods**                                   |                           |                                                                                                                                                                  |
 | `Tables.getcolumn(row, ::Type{T}, i::Int, nm::Symbol)` | Tables.getcolumn(row, nm) | Given a column element type `T`, index `i`, and column name `nm`, retrieve the column value. Provides a type-stable or even constant-prop-able mechanism for efficiency. |
-| `Tables.ncol(row)`                                    | length(propertynames(row) | Return number of columns                                                                                                                                         |
+| `Tables.ncol(row)`                                     | length(propertynames(row) | Return number of columns                                                                                                                                         |
 
 Note that subtypes of `Tables.AbstractRow` **must** overload all required methods listed above
 instead of relying on these methods' default definitions.
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -208,7 +208,7 @@ function Base.show(io::IO, x::T) where {T <: AbstractRow}
 end
 
 function Base.show(io::IO, table::AbstractColumns; max_cols = 20)
-    print(io, "$(typeof(table)) with $(rowcount(table)) rows, $(columncount(table)) columns, and ")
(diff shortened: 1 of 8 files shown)


bkamins:
The DataAPI.jl defines `nrow` and `ncol` as a public API for this. `rowcount` is internal. If we are at it it would be probably better to properly implement default `nrow` and `ncol` implementations in this PR.

rikhuijzer:
> If we are at it it would be probably better to properly implement default `nrow` and `ncol` implementations in this PR.

That would be great yes. Then `Tables.jl` sources can provide more efficient methods.

> `rowcount` is internal.

How can I make `nrow` and `ncol` public for `Tables.jl`? I'm not sure what the convention is here

bkamins (approved):
Now I see that what I wanted is a minimal type piracy, but I think it is OK.

bkamins:
> How can I make `nrow` and `ncol` public for Tables.jl? I'm not sure what the convention is here

You do not need to do anything more than you did. Tables.jl does not export anything, but `nrow` and `ncol` are public because they implement public interface defined in DataAPI.jl.

src/fallbacks.jl
```diff
@@ -4,12 +4,16 @@
 
 # Turn any AbstractColumns into an AbstractRow iterator
 
-# get the number of rows in the incoming table
-function rowcount(cols)
+"Return the number of rows in the incoming table."
```
quinnj:
```suggestion
"""
    nrow(::AbstractColumns)
    
Returns the number of rows in an object that satisfies the `AbstractColumns` interface
as returned from `Tables.columns(tbl)`. Note that this isn't valid to call on _any_ valid
Tables.jl source as row-oriented tables may not have a defined length.
"""
```

src/fallbacks.jl
```diff
@@ -4,12 +4,16 @@
...
+function nrow(cols)
```
quinnj: @bkamins, does this not pirate a default definition? I'm a little uncomfortable with this definition because technically `rowcount` was only meant to be called on the result of `Tables.columns`, but now it "looks like" it works generically on any table. And for the most part, it probably will, but we're not really following the principles of the interface.

src/fallbacks.jl
```diff
@@ -4,12 +4,16 @@
...
     names = columnnames(cols)
     isempty(names) && return 0
     return length(getcolumn(cols, names[1]))
 end
+@deprecate rowcount(x) nrow(x)
+
+"Return the number of columns in the incoming table."
+ncol(cols) = length(columnnames(cols))
```
quinnj: Same here; `columnnames` and `getcolumn` are only technically allowed to be called on the result of `Tables.columns` or on individual iterated rows from `Tables.rows`, so this makes me a bit uncomfortable.

src/fallbacks.jl
bkamins: > does this not pirate a default definition?

DataAPI.jl does not define any method for this function. Same for `ncol` - DataAPI.jl does not define any method for this function.

As I have commented - it does pirate as it add the definition to `ncol` with `::Any` signature. The problem is exactly what you have commented - some tables might not have a well defined row count (maybe also the same problem is for `ncol`). So what do you think would be best to do? What I would ideally have is some definition that works by default correctly. I think that if default errors it is not a huge problem if it does not error on "standard" tables.

Maybe what we should do is:
* leave `rowcount` to be internal;
* define `nrow` and `ncol` only for tables defined in Tables.jl tables (`rowtable`, `coltable`, `dictrowtable`, `dictcolumntable`, `table` maybe some more? - you know best)

quinnj: Yes, that's what I would prefer (only define for Tables.jl-defined tables. I'm also happy to add recommendations to the docs that table types should strongly consider implementing `nrow`/`ncol` on their own table types as apart of the API.

rikhuijzer:
@bkamins @quinnj Is 5722256 going in the expected direction? Probably needs some more tests and I probably should implement a few more methods?

bkamins:
@quinnj will know best if you correctly covered everything :), but for matrices things look good :).

src/Tables.jl
```diff
@@ -32,6 +32,7 @@ Interface definition:
 | `Tables.columnnames(table)`                              | propertynames(table)        | Return column names for a table as an indexable collection                                                                                                   |
 | **Optional methods**                                     |                             |                                                                                                                                                              |
 | `Tables.getcolumn(table, ::Type{T}, i::Int, nm::Symbol)` | Tables.getcolumn(table, nm) | Given a column eltype `T`, index `i`, and column name `nm`, retrieve the column. Provides a type-stable or even constant-prop-able mechanism for efficiency. |
+| `Tables.ncols(table)`                                    | Tables.ncol(table)          | Return the number of columns                                                                                                                                 |
```
CarloLucibello:
```suggestion
| `Tables.ncol(table)`                                    | Tables.ncol(table)          | Return the number of columns                                                                                                                                 |
```
and the same change also in other spots

rikhuijzer: Well spotted. Fixed in f094263

quinnj:
Sorry this sat to rot so long; I believe it was fully implemented in #343

PR #292, opened 2022-09-06
Change getrows -> subset

quinnj:
Also adds a generic implementation of subset for column-oriented tables
and row-oriented tables that are AbstractVector. Adds implementation and
tests for DictRowTable/DictColumnTable.

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -569,15 +569,15 @@ struct Partitioner{T}
 end
 
 """
-    Tables.getrows(x, inds; view=nothing)
+    Tables.subset(x, inds; view=nothing)
 
 Return one or more rows from table `x` according to the position(s) specified by `inds`:
 
 - If `inds` is a single non-boolean integer return a row object.
-- If `inds` is a vector of non-boolean integers, a vector of booleans, or a `:`, return an indexable object of rows. 
+- If `inds` is a vector of non-boolean integers, a vector of booleans, or a `:`, return a subset of the original table according to the indices.
   In this case, the returned type is not necessarily the same as the original table type.
   
-If other type of `inds` is passed than specified above the behavior is undefined.
+If other types of `inds` are passed than specified above the behavior is undefined.
 
 The `view` argument influences whether the returned object is a view of the original table
 or an independent copy:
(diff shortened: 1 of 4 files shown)


src/Tables.jl
```diff
@@ -587,11 +587,37 @@ or an independent copy:
...
 Support for `view=true` or `view=false` is optional
 (i.e. implementations might error on them if they are not supported).
 """
-function getrows end
+function subset(x::T, inds; view::Union{Bool, Nothing}=nothing) where {T}
+    # because this method is being called, we know `x` didn't define it's own Tables.subset
+    # first check if it supports column access, and if so, apply inds and wrap columns in a DictColumnTable
+    if columnaccess(x)
+        cols = columns(x)
+        if inds isa Integer
```
bkamins, line 599: `Bool` should be excluded here and throw an error. (or maybe `ColumnsRow(cols, inds)` already does it?)

src/Tables.jl
```diff
@@ -587,11 +587,37 @@ or an independent copy:
...
+            return ColumnsRow(cols, inds)
+        else
+            ret = view === true ? _map(c -> Base.view(c, inds), cols) : _map(c -> c[inds], cols)
+            return DictColumnTable(schema(cols), ret)
```
bkamins, line 603: does `DictColumnTable` check that it gets vectors as columns?

bkamins, line 603: As a special case someone can pass `CartesianIndex` as `inds` - do we want to handle it in the `ColumnsRow` case or error on it here?

src/dicts.jl
```diff
@@ -122,6 +124,16 @@ function Base.iterate(x::DictRowTable, st=1)
     return DictRow(x.names, x.values[st]), st + 1
 end
 
+function subset(x::DictRowTable, inds; view::Union{Bool,Nothing} = nothing)
```
bkamins, line 127: similar comments as above to `inds` apply here

src/Tables.jl
quinnj: Ah, great catch.

quinnj: It is already covered

src/Tables.jl
quinnj: Looks like our existing `vectorcheck` handles that

ablaom:
https://github.com/JuliaData/Tables.jl/pull/284#issuecomment-1238687508

Issue #287, opened 2022-08-23
Might Tables.jl be a new Home for `StatsPlots.@df` ?


BeastyBlacksmith:
The `@df` macro allows writing column names in function calls as in 
```julia
using StatsPlots
import RDatasets
iris = RDatasets.dataset("datasets", "iris")
@df iris scatter(:SepalLength, :SepalWidth, group = :Species)
@df iris sum(:SepalLength)  # not only useful for plotting
```
It will be removed from StatsPlots.jl probably in 1.0 and if we don't find a better home, it will got to `Plots.jl` (https://github.com/JuliaPlots/Plots.jl/pull/3351), but actually I think it should be somewhere more general purpose.

If it's going here it should be probably be renamed to something like `@table`.

rofinn:
I don't think Tables.jl is the right place for it. Maybe we should just have a TablePlots.jl package similar to how we have TableOperations.jl?

BeastyBlacksmith:
> Maybe we should just have a TablePlots.jl package

That doesn't make sense, since this macro is not tied to plotting at all. It works for any function call/ block of function calls. It's only historic accident that it ended up in StatsPlots.jl, since there was no Table interface back then.

The only thing that is kind of plotting related is the automatic passing of a `label` keyword if, the function to be called supports this, but that part could also be stripped and put in a Plots-specific extension of this macro.

nalimilan:
This sounds like something which could live in DataFramesMeta (with a different name). Cc: @pdeffebach @bkamins

pdeffebach:
Can someone try and see how much of the `@df` macro functionality can be replicated by just using `@with` from DataFramesMeta.jl?

bkamins:
I already checked it. `@with` is better. The only problem is that `@with` works with data frames, while `@df` works with any tables.

pdeffebach:
Ah. 

Current implementation of `@with` does not use any DataFrames.jl-specific things actually, so I could make it broader.

bkamins:
So if we document that `@with` contract is Tables.jl (which I think should be easy) then `@df` can be made an alias for `@with` if someone wants it.

Other than that `@with` has all functionality that `@df` has but it has some more functionality for better handling special cases (like `$`)

BeastyBlacksmith:
Thats a good idea. Could we then move `@with` to Tables.jl?

nalimilan:
[...]

PR #293, opened 2022-09-07
add ByRow

bkamins:
Following the discussion in https://github.com/JuliaData/Tables.jl/issues/287.

CC @quinnj @nalimilan @pdeffebach

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -168,7 +168,7 @@ _row_to_named_tuple(row) = NamedTuple(Row(row))
 
 `ByRow(f)` can be passed two types of arguments:
 - One or more 1-based `AbstractVector`s of equal length: In this case the returned value
-is a vector resulting from applying `f` to elements of passed vectors element-wise. 
+is a vector resulting from applying `f` to elements of passed vectors element-wise.
 Function `f` is called exactly once for each element of passed vectors (as opposed to `map`
 which assumes for some types of source vectors (e.g. `SparseVector`) that the
 wrapped function is pure, and may call the function `f` only once for multiple
@@ -189,24 +189,26 @@ source table).
 
 # Examples
 ```
-julia> ByRow(x -> x^2)(1:3)
+julia> Tables.ByRow(x -> x^2)(1:3)
 3-element Vector{Int64}:
  1
  4
  9
 
-julia> ByRow((x, y) -> x*y)(1:3, 2:4)
+julia> Tables.ByRow((x, y) -> x*y)(1:3, 2:4)
 3-element Vector{Int64}:
   2
   6
  12
 
-julia> ByRow(x -> x.a)((a=1:2, b=3:4))
+julia> Tables.ByRow(x -> x.a)((a=1:2, b=3:4))
 2-element Vector{Int64}:
  1
  2
- 
- julia> ByRow(x -> (a=x.a*2, b=sin(x.b), c=x.c))((a=[1,2,3], b=[1.2, 3.4, 5.6], c=["a","b","c"]))
+
+ julia> Tables.ByRow(x -> (a=x.a*2, b=sin(x.b), c=x.c))((a=[1, 2, 3],
+                                                         b=[1.2, 3.4, 5.6],
+                                                         c=["a", "b", "c"]))
 3-element Vector{NamedTuple{(:a, :b, :c), Tuple{Int64, Float64, String}}}:
  (a = 2, b = 0.9320390859672263, c = "a")
  (a = 4, b = -0.2555411020268312, c = "b")
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -205,6 +205,13 @@ julia> ByRow(x -> x.a)((a=1:2, b=3:4))
 2-element Vector{Int64}:
  1
  2
+ 
+ julia> ByRow(x -> (a=x.a*2, b=sin(x.b), c=x.c))((a=[1,2,3], b=[1.2, 3.4, 5.6], c=["a","b","c"]))
+3-element Vector{NamedTuple{(:a, :b, :c), Tuple{Int64, Float64, String}}}:
+ (a = 2, b = 0.9320390859672263, c = "a")
+ (a = 4, b = -0.2555411020268312, c = "b")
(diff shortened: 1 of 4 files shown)


bkamins:
@quinnj I have also made some small clean-up proposals to src/namedtuples.jl - I hope they are correct 😄.

src/utils.jl
```diff
@@ -160,3 +160,81 @@ rowmerge(row; fields_to_merge...) = rowmerge(row, values(fields_to_merge))
 
 _row_to_named_tuple(row::NamedTuple) = row
 _row_to_named_tuple(row) = NamedTuple(Row(row))
+
+"""
+    ByRow
+
+`ByRow(f)` is a function that is wrapper around function `f` that vectorizes it.
```
nalimilan: More explicit than "vectorize"?
```suggestion
`ByRow(f)` returns a function which applies function `f` to each element in a vector.
```

src/utils.jl
```diff
@@ -160,3 +160,81 @@ rowmerge(row; fields_to_merge...) = rowmerge(row, values(fields_to_merge))
 
 _row_to_named_tuple(row::NamedTuple) = row
 _row_to_named_tuple(row) = NamedTuple(Row(row))
+
+"""
+    ByRow
```
nalimilan:
```suggestion
    ByRow <: Function
```

src/utils.jl
```diff
@@ -160,3 +160,81 @@ rowmerge(row; fields_to_merge...) = rowmerge(row, values(fields_to_merge))
...
+
+As a special case `ByRow(f)` also accepts passing a `Tables.ColumnTable` as its
+argument. In this case the function `f` is passed a `NamedTuple` that is created
+from rows of passed table.
+
+The return value of `ByRow` is always a vector.
+
+Vectors passed to `ByRow(f)` must have equal length and use 1-based indexing.
+Also vectors stored in the `Tables.ColumnTable` if it is passed to `ByRow(f)`
+must meet this condition.
```
nalimilan:
```suggestion
`ByRow(f)` can be passed two types of arguments:
- One or more 1-based `AbstractVector`s of equal length: In this case the returned value
is a vector resulting from applying `f` to elements of passed vectors element-wise. 
Function `f` is called exactly once for each element of passed vectors (as opposed to `map`
which assumes for some types of source vectors (e.g. `SparseVector`) that the
wrapped function is pure, and may call the function `f` only once for multiple
equal values.
- A `Tables.ColumnTable` holding 1-based columns of equal length: In this case the function
`f` is passed a `NamedTuple` created for each row of passed table.

The return value of `ByRow(f)` is always a vector.
```

test/runtests.jl
```diff
@@ -846,3 +846,36 @@ Tables.columnnames(::WideTable2) = [Symbol("x", i) for i = 1:1000]
...
+    @test Tables.ByRow(x -> x^2)(1:3) == [1, 4, 9]
+    @test Tables.ByRow((x, y) -> x*y)(1:3, 2:4) == [2, 6, 12]
+    @test Tables.ByRow((x, y, z) -> x*y*z)(1:3, 2:4, 3:5) == [6, 24, 60]
+    let i = 0
+        @test Tables.ByRow(x -> (i += 1; return i))(spzeros(5)) == 1:5
+    end
+    # check that map is not what we want
+    let i = 0
+        @test map(x -> (i += 1; return i), spzeros(5)) == [1, 1, 1, 1, 1]
+    end
```
nalimilan, line 860: Do we really need to test this? We don't really care how `map` is implemented in SparseArrays.

nalimilan (approved):

test/runtests.jl
bkamins: We do not need it, but I think it is not a problem and shows the reader of tests the difference (which is not obvious I would say).

src/utils.jl
```diff
@@ -160,3 +160,75 @@ rowmerge(row; fields_to_merge...) = rowmerge(row, values(fields_to_merge))
 
 _row_to_named_tuple(row::NamedTuple) = row
 _row_to_named_tuple(row) = NamedTuple(Row(row))
+
+"""
+    ByRow <: Function
+
+`ByRow(f)` returns a function which applies function `f` to each element in a vector.
+
+`ByRow(f)` can be passed two types of arguments:
```
bkamins, line 169: @nalimilan also being 1-based is a requirement.

bkamins:
@quinnj and @pdeffebach - can you please check and approve as this is a surprisingly complex PR (fortunately not big)

src/utils.jl
```diff
@@ -160,3 +160,75 @@ rowmerge(row; fields_to_merge...) = rowmerge(row, values(fields_to_merge))
...
+julia> ByRow((x, y) -> x*y)(1:3, 2:4)
+3-element Vector{Int64}:
+  2
+  6
+ 12
+
+julia> ByRow(x -> x.a)((a=1:2, b=3:4))
+2-element Vector{Int64}:
+ 1
+ 2
```
quinnj, line 207:
```suggestion
 2
 
 julia> ByRow(x -> (a=x.a*2, b=sin(x.b), c=x.c))((a=[1,2,3], b=[1.2, 3.4, 5.6], c=["a","b","c"]))
3-element Vector{NamedTuple{(:a, :b, :c), Tuple{Int64, Float64, String}}}:
 (a = 2, b = 0.9320390859672263, c = "a")
 (a = 4, b = -0.2555411020268312, c = "b")
 (a = 6, b = -0.6312666378723216, c = "c")

```

quinnj: I think it'd be nice to have a namedtuple -> namedtuple example here. I tend to prefer in documentation to include package prefix when the item isn't exported; would you mind changing all the usages here to `Tables.ByRow`, so it's extra clear?

quinnj (approved):
Sorry for the slow review, but this LGTM. I left one comment on an example doc + doc package prefix.

bkamins:
Fixed. The `Tables.` prefix for sure should be there. I had DataFrames.jl loaded, which exports it so I forgotten to add it in the docstring.

PR #298, opened 2022-09-16
Fix allocatecolumn

bkamins:
The reason for the fix is that old code:
* in case eltype is `Any` used `undef` branch (which most of the time works, but it is not guaranteed)
* not all array types support initializer with a value other than `undef` (e.g. this is the case for `CategoricalArray`)

What I propose is a safe approach, but maybe there would be better ideas.

x-ref: massive errors in DataFrames.jl https://github.com/JuliaData/DataFrames.jl/pull/3158

CC @quinnj @nalimilan

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -103,6 +103,7 @@ end
 
 Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
 Alternatively, and more generally, custom scalars can overload `DataAPI.defaultarray` to signal the default array type.
+In this case the signaled array type must support a constructor accepting `undef` for initialization.
 """
 function allocatecolumn(T, len)
     a = DataAPI.defaultarray(T, 1)(undef, len)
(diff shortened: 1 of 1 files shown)


src/fallbacks.jl
```diff
@@ -104,7 +104,11 @@ end
 Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
 Alternatively, and more generally, custom scalars can overload `DataAPI.defaultarray` to signal the default array type.
 """
-allocatecolumn(T, len) = DataAPI.defaultarray(T, 1)(Base.nonmissingtype(T) === T ? undef : missing, len)
+function allocatecolumn(T, len)
+    a = DataAPI.defaultarray(T, 1)(Base.nonmissingtype(T) === T ? undef : missing, len)
```
bkamins:
```suggestion
    a = DataAPI.defaultarray(T, 1)(undef, len)
```

src/fallbacks.jl
```diff
@@ -104,7 +104,11 @@ end
 Custom column types can override with an appropriate "scalar" element type that should dispatch to their column allocator.
 Alternatively, and more generally, custom scalars can overload `DataAPI.defaultarray` to signal the default array type.
```
bkamins, line 105:
```suggestion
Alternatively, and more generally, custom scalars can overload `DataAPI.defaultarray` to signal the default array type.
In this case the signaled array type must support a constructor accepting `undef` for initialization.
```

quinnj (approved):

bkamins:
Thank you! Can we have a patch release?

PR #300, opened 2022-09-17
Fix Tables.subset implementation for rows and view

quinnj:


bkamins:
The issue with view is more complex than I thought. Maybe we should the sentence:

> Support for `view=true` or `view=false` is optional (i.e. implementations might error on them if they are not supported).

to

> Support for `view=true` or `view=false` is optional; implementations may respect the value passed in the `view` keyword argument or ignore it

The point is, after thinking about it:
* throwing an error is not useful in practice
* ensuring that `view` is correctly handled for all tables may be hard

bkamins:
Do you think we should change the docstring or not?

quinnj:
Clarified in https://github.com/JuliaData/Tables.jl/commit/63067e3d6ec445ea8021b4460dd0820679acd16c

PR #304, opened 2022-09-21
Change view to viewhint in Tables.subset

### LINKED ISSUE #303: View hint
Currently `subset` takes `view=` which is a hint that may be ignored. I don't like the idea of ignoring keyword argument values, so I'm considering the issue. On Slack some approaches were discussed, 

One option is to rename `view` to `viewhint`. 

Another option is the `view` could be a hint supplemented by a second argument `force` or `strict`. I normally prefer that `force`/`strict` arguments be explicit about what exactly is being forced, for example `forceviewoption`. 

But then once that's the situation, it seems odd to have both `view` and an option for saying "I really mean it" rather than just passing either `viewhint` or `view` with the latter being enforced. The enforced `view` argument could be added in a separate PR, which would make this option very similar to the first.

Therefore I propose renaming `view` to `viewhint`. Then in the future if users so desire, add a `view` argument that would be enforced. They would combine according to the following rules:

```jl
@match (view, viewhint) begin
(true, _) => enforce true
(false, _) => enforce false
(nothing, true) => prefer true
(nothing, false) => prefer false
end
```

An alternative future implementation could define a `subsetview` function if desired.

quinnj:


bkamins (approved):
Looks good.

@quinnj - do you think https://tables.juliadata.org/dev/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source) and https://tables.juliadata.org/dev/#Using-the-Interface-(i.e.-consuming-Tables.jl-compatible-sources) sections in the manual should be updated to mention `Tables.subset`?

bkamins:
@jariji and @ablaom - can you have a quick look at the PR? Thank you!

jariji:
LGTM, cheers.

ablaom (approved):
Looks good to me, thank you.

bkamins:
@quinnj - do we make this change and make a release (@nalimilan and I need to know it before releasing DataFrames.jl 1.4 😄)

quinnj:
Merged and tagged a 1.9 release

PR #307, opened 2022-10-09
Fix docstring for Tables.rows

ablaom:
The present docstring for `Tables.rows` suggests that `getcolumn` can be applied to `Tables.rows(X)`, whereas the contract is only that it can be applied to a single row of that object, right?


quinnj (approved):
Good catch, thanks!

Issue #316, opened 2022-11-13
Tables that subtype AbstractMatrix are not fully supported


LilithHafner:
I have a type that subtypes AbstractMatrix _and_ conforms to the Tables.table interface (I think). Yet `Tables.rows(::MyType)` errors, dispatching to https://github.com/JuliaData/Tables.jl/blob/db5e0903cf40a0803ec23ebd92753b70feeab169/src/matrix.jl#L3 without ever checking if my table conforms to the Tables.table interface.

<details>
<summary>Here's a large MWE, I can provide a smaller one upon request</summary>

```julia
using AxisArrays, Tables

abstract type AbstractMatrixTable{T} <: AbstractMatrix{T} end

struct MatrixTable{T, U <: AxisArray{T, 2}} <: AbstractMatrixTable{T}
    data::U
    MatrixTable(data::AxisArray{T, 2}) where T = new{T, typeof(data)}(data)
end
function MatrixTable(data::AbstractMatrix, rows::AbstractVector, cols::Vector{Symbol}; check_unique::Bool=true)
    check_unique && !allunique(cols) && throw(ArgumentError("Column names must be unique"))
    MatrixTable(AxisArray(data; rows, cols))
end

# getter method to avoid conflict with custom getproperty method
data(m::MatrixTable) = getfield(m, :data)

# named axes
named_axes(m::AbstractMatrixTable) = getproperty.(data(m).axes, :val)
rows(m::AbstractMatrixTable) = named_axes(m)[1]
cols(m::AbstractMatrixTable) = named_axes(m)[2]

# Access by dot syntax `data.var3`
Base.propertynames(m::AbstractMatrixTable) = cols(m)
Base.getproperty(m::AbstractMatrixTable, col::Symbol) = getindex(m, :, col)

# Conform to the AbstractMatrix interface
# https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array

# Required
Base.size(m::AbstractMatrixTable) = size(data(m))
function Base.getindex(m::AbstractMatrixTable, inds...) # Access keeps the same type (uncommon design decision); with and exception for m[today(),:b] === 7
    a = data(m)
    ix = AxisArrays.to_index(a, inds...)
    all(i -> i isa Integer, ix) && return AxisArrays.getindex_converted(a, ix...)
    MatrixTable(AxisArrays.getindex_converted(a, (i isa Integer ? (i:i) : i for i in ix)...))
end
[...]

PR #317, opened 2022-11-13
Support tables that subtype AbstractMatrix

LilithHafner:
Support tables that subtype AbstractMatrix and define only one of `rows` or `columns`. I copied the implementations from fallbacks.jl.

Fixes #316

PR #319, opened 2022-11-29
If AbstractMatrix defines Tables APIs, make fallbacks work

quinnj:
Alternative fix for #316.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -82,7 +82,10 @@ end
 # this is our generic Tables.rows fallback definition
 @noinline nodefault(T) = throw(ArgumentError("no default `Tables.rows` implementation for type: $T"))
 
-function rows(x::T) where {T}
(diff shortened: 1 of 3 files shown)


src/fallbacks.jl
```diff
@@ -252,7 +255,9 @@ getcolumn(x::CopiedColumns, nm::Symbol) = getcolumn(source(x), nm)
 columnnames(x::CopiedColumns) = columnnames(source(x))
 
 # here's our generic fallback Tables.columns definition
-@inline function columns(x::T) where {T}
+columns(x::T) where {T} = _columns(x)
```
LilithHafner:
```suggestion
@inline columns(x::T) where {T} = _columns(x)
```
Keep the inline annotation?

LilithHafner (commented):
This is a better approach than #317 and includes tests. I support it.

Issue #322, opened 2023-01-22
Calling `Tables.istable` is too slow if it's not actually a table


rafaqz:
Calling `Tables.istable`  ends up at `hasmethod` as a fallback. (via TableTraits.jl and IteratorInterfaceExtensions.jl)

This is super slow. It can be seen as the big chunk in blue/green in this flame graph. The rest of the graph is rasterizing a whole vector of polygons, which in this case is not a table.

![2023-01-22-224421_1920x1080](https://user-images.githubusercontent.com/2534009/213941949-fb6751f2-cb67-4427-817c-4e0c2bd19f6b.png)

Is there a way to check if something is a table without this fallback?

The use case here is the object can be a table with the target column (geometries) and some extra columns we may use, or just some iterator of geometries. I dont know a way to separate these other than `istable` - but the iterators get this overhead because they are not tables.

rafaqz:
The problem here seems to be that I'm iterating over an object and  each of its contents is passed to a method that again checks `istable`. 

~It benchmarks at 160ns per call, which is not much at all if you just do it once. Its just too much to be used as a guard when iterating over lots of things. I had assumed it was a type level check so it would essentially be free.~

Actually `istable` was taking 40 μs per call inside the function! I'm not sure why it benchmarks faster in the REPL or what the interaction is.


My solution is to use GeoInterface.jl traits to filter objects first, because they are compile time traits. Some iterators will still be slow. It seems that if multiple packages had traits that check if methods exist this would start to be a problem.


Some benchmarks (of `rasterize` in Rasters.jl). This run has 4 calls to `istable`, always `false`:
```julia
julia> @benchmark rasterize(sum, $polygons; res=5, fill=1, boundary=:center)
BenchmarkTools.Trial: 8682 samples with 1 evaluation.
 Range (min … max):  489.019 μs … 109.038 ms  ┊ GC (min … max): 0.00% … 69.94%
 Time  (median):     513.876 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   578.621 μs ±   2.509 ms  ┊ GC (mean ± σ):  7.34% ±  1.68%

                        ▁▆█▅▂                     ▂▄             
  ▁▁▂▂▂▂▂▂▂▂▂▂▁▂▂▂▁▁▁▁▁▃█████▅▄▃▄▅▅▄▃▃▂▃▃▃▃▃▃▂▂▂▂▅██▇▃▂▂▂▂▃▂▂▁▁ ▃
[...]

PR #323, opened 2023-01-24
document isiterabletable fallback

rafaqz:
This partially solves #322 by documenting the fallback behavior of `istable` and the workaround for the slow runtime performance when called on an object that is not known to be a table at compile time.


quinnj (approved):
Thanks!

PR #329, opened 2023-03-13
eltype should take type rather than instance

baumgold:


quinnj:
Looks like this is breaking some of the tests; not sure if the tests also should just be updated to check the type instead of instance (probably), or if there are cases where we're actually relying on knowing the eltype from the instance instead of just the type (which we probably shouldn't do...).

baumgold:
> Looks like this is breaking some of the tests; not sure if the tests also should just be updated to check the type instead of instance (probably), or if there are cases where we're actually relying on knowing the eltype from the instance instead of just the type (which we probably shouldn't do...).

My mistake.  I created the PR too hastily.  The problem is that the first implementation below is broken while the second one works.  I fixed the implementations everywhere and now all tests pass.

```julia
Base.eltype(::Type{DataValueRowIterator{NT}}) where {NT} = NT
```

```julia
Base.eltype(::Type{D}) where {NT, D<:DataValueRowIterator{NT}} = NT
```

quinnj (approved):
Thanks; I tried to check all those cases, but got lazy and didn't check the last one!

PR #334, opened 2023-04-23
Use `DataAPI.rownumber`

VEZY:
Now using the newly proposed `DataAPI.rownumber` interface (https://github.com/JuliaData/DataAPI.jl/issues/60). We add an implementation for: `Tables.ColumnsRow`, `Tables.DictRow`, `Tables.IteratorRow` and `Tables.MatrixRow`.

Note: a similar function (`getrow`) existed already for `ColumnsRow` and `IteratorRow`. These functions were not removed as a precaution, but rather used for implementing `DataAPI.rownumber` (*i.e.* `rownumber(c::ColumnsRow) = getrow(c)`)

The implementation for `Tables.Row` is just `rownumber(x::Row) = rownumber(x.x)`. Maybe we should add a fallback or throw an explicit error if the row type does not implement `rownumber` ?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.10.1"
+version = "1.11.0"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ TableTraits = "3783bdb8-4a98-5b6b-af9a-565f29a5fe9c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
-DataAPI = "1"
+DataAPI = "1.15"
 DataValueInterfaces = "1"
 IteratorInterfaceExtensions = "0.1.1, 1"
 OrderedCollections = "1"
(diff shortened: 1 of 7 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.10.1"
```
bkamins, line 4: Also DataAPI.jl version requirement needs to be changed below.

bkamins:
Can you please add tests for all 5 types?

Project.toml
VEZY: Ho yes I forgot this you're right. Done in aefedf63f805ce6d8485229d678893dbc4b1132c

VEZY:
Hoo, when making the tests I did realise I was completely misinterpreting the row argument, it was not an index, but the actual data of the row... That's why tests are so valuable ^^.

So I don't think we can get the row number without adding it as a field of the rows structures for: `Tables.ColumnsRow`, `Tables.DictRow`, `Tables.IteratorRow` and `Tables.MatrixRow`.

There's two choices from here, either we add a field for the row index into these structures, or we don't define `DataAPI.rownumber` for these.

I made an example implementation of a row number in `DictRow` (5f091b0). If you think that's a good path forward, I'll implement it the same way for the others.

bkamins:
> There's two choices from here, either we add a field for the row index into these structures, or we don't define `DataAPI.rownumber` for these.

Let us wait for @quinnj to decide on this. My thinking is that:
* it should not add performance overhead
* having this information can be useful for users

so I would be OK to add it.

The only downside could be that in some cases these objects could be generated from sources that do not have a proper row numbering. However, I am not sure if this is possible. @quinnj should know.

quinnj:
Sorry this was never followed up on; closing as stale/old. Feel free to ping/comment if something is still needed.

Issue #332, opened 2023-04-10
Implement `DataAPI.nrow` for in-memory table types owned by Tables.jl


ablaom:
These include `MatrixTable`, `DictRowTable`, `DictColumnTable`, but perhaps there are others I don't know about. 

In support of: #330

EssamWisam:
I plan to shortly make a PR for this.

quinnj:
Implemented by #343, including matrix, dictionary, row-table, and column-table types.

PR #336, opened 2023-06-23
Implement DataAPI.nrow for MatrixTable, DictColumnTable and DictColumnTable

EssamWisam:
In response to issue #332


bkamins:
maybe adding support of `ncol` also would make sense in the same PR?

EssamWisam:
> maybe adding support of `ncol` also would make sense in the same PR?

Indeed, added that.

bkamins (approved):

bkamins:
@quinnj - these implementations look good to me. They use internal fields of structs so could you please have a look at them?

PR #340, opened 2023-08-05
Fix type instability in `getproperty(::Schema, ::Symbol)`

### LINKED ISSUE #339: Type stability of `Tables.matrix(::NamedTuple)`
I'm descending some type instability issues in my package and realized that `Tables.matrix(::NamedTuple)` is not type stable. Would this be feasible to implement?

For example:

```julia
julia> using Tables

julia> X = (x=randn(32), y=randn(32));

julia> typeof(Tables.matrix(X))
Matrix{Float64} (alias for Array{Float64, 2})

julia> @code_warntype Tables.matrix(X)
MethodInstance for Tables.matrix(::@NamedTuple{x::Vector{Float64}, y::Vector{Float64}})
  from matrix(table; transpose) @ Tables ~/.julia/packages/Tables/AcRIE/src/matrix.jl:84
Arguments
  #self#::Core.Const(Tables.matrix)
  table::@NamedTuple{x::Vector{Float64}, y::Vector{Float64}}
Body::Matrix
1 ─ %1 = Tables.:(var"#matrix#85")(false, #self#, table)::Matrix
└──      return %1
```

So the type inference is not able to infer that this is a `Matrix{Float64}`, only that it is a `Matrix{Any}`, despite the method call being specialized to a NamedTuple with `Vector{Float64}` for columns. 

Version info:

```
Julia Version 1.10.0-beta1
Commit 6616549950e (2023-07-25 17:43 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M1 Pro
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
  Threads: 8 on 6 virtual cores
Environment:
  JULIA_FORMATTER_SO = /Users/mcranmer/julia_formatter.so
  JULIA_NUM_THREADS = auto
  JULIA_EDITOR = code
```

MilesCranmer:
Fixes #339. The issue was that inside `getproperty`, the generator for table types:

```julia
Tuple(fieldtype(types, i) for i = 1:fieldcount(types))
```

is type unstable.

This PR changes it to:

```julia
ntuple(i -> fieldtype(types, i), Val(fieldcount(types)))
```

which is stable.

**Edit:** 4d4c9621cdb38d856301a473218b5e3a5a8de974 changes it so that `getproperty` falls back to the type unstable `Tuple` when the number of columns goes above 512. This prevents any major slow downs in compile time.

This gives some major speed ups in both runtime and compile time:

plot

e.g., here is an inspection of `@code_warntype` on a function that returns `.types` for a schema:

```julia
julia> X = (x=randn(32), y=randn(32));

julia> f(X) = Tables.schema(X).types
f (generic function with 1 method)

julia> @code_warntype f(X)
MethodInstance for f(::@NamedTuple{x::Vector{Float64}, y::Vector{Float64}})
  from f(X) @ Main REPL[5]:1
Arguments
  #self#::Core.Const(f)
  X::@NamedTuple{x::Vector{Float64}, y::Vector{Float64}}
Body::Tuple{DataType, DataType}
1 ─ %1 = Tables.schema::Core.Const(Tables.schema)
│   %2 = (%1)(X)::Core.Const(Tables.Schema:
 :x  Float64
 :y  Float64)
│   %3 = Base.getproperty(%2, :types)::Core.Const((Float64, Float64))
└──      return %3
```

whereas before this change, it was inferring this as `Vararg{DataType}` (bad).

@quinnj @bkamins could you please review and merge when you get a chance? This is necessary to fix a variety of type instabilities in MLJ interfaces.

cc @ablaom @OkonSamuel. This should speed up calls to `fit` and `predict` across the MLJ ecosystem as it removes a key type instability.

Thanks!
Miles

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -499,8 +499,8 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
     if field === :names
         return names === nothing ? getfield(sch, :storednames) : names
     elseif field === :types
-        T = getfield(sch, :storedtypes)
         if types === nothing
+            T = getfield(sch, :storedtypes)
             return (T !== nothing ? T : nothing)
         else
             ncol = fieldcount(types)
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -500,8 +500,7 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
         return names === nothing ? getfield(sch, :storednames) : names
     elseif field === :types
         if types === nothing
-            T = getfield(sch, :storedtypes)
-            return T !== nothing ? T : nothing
+            return getfield(sch, :storedtypes)
         else
             ncol = fieldcount(types)
             if ncol <= 512
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -500,7 +500,7 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
         return names === nothing ? getfield(sch, :storednames) : names
     elseif field === :types
         T = getfield(sch, :storedtypes)
-        return types === nothing ? (T !== nothing ? T : nothing) : Tuple(fieldtype(types, i) for i = 1:fieldcount(types))
+        return types === nothing ? (T !== nothing ? T : nothing) : ntuple(i -> fieldtype(types, i), Val(fieldcount(types)))
     else
         throw(ArgumentError("unsupported property for Tables.Schema"))
     end
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -501,7 +501,7 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
     elseif field === :types
         if types === nothing
             T = getfield(sch, :storedtypes)
-            return (T !== nothing ? T : nothing)
+            return T !== nothing ? T : nothing
         else
             ncol = fieldcount(types)
             if ncol <= 512
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -500,7 +500,17 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
         return names === nothing ? getfield(sch, :storednames) : names
     elseif field === :types
         T = getfield(sch, :storedtypes)
-        return types === nothing ? (T !== nothing ? T : nothing) : ntuple(i -> fieldtype(types, i), Val(fieldcount(types)))
+        if types === nothing
+            return (T !== nothing ? T : nothing)
+        else
+            ncol = fieldcount(types)
+            if ncol <= 512
+                # Type stable, but slower to compile
+                return ntuple(i -> fieldtype(types, i), Val(ncol))
+            else
+                return Tuple(fieldtype(types, i) for i=1:ncol)
+            end
+        end
     else
         throw(ArgumentError("unsupported property for Tables.Schema"))
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -214,8 +214,13 @@ end
 
     @testset "Type stability of schema(x).types" begin
         _get_types(X) = Tables.schema(X).types
+
         x = (a=[1.0], b=[1.0])
-        @inferred _get_types(x)
+        @inferred Tuple{Type{Float64},Type{Float64}} _get_types(x)
+
+        # Trigger other branch which lacks type stability:
+        x_wide = NamedTuple([Symbol("x$(i)") => [1.0] for i=1:513])
+        @inferred NTuple{513,DataType} _get_types(x_wide)
     end
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -211,6 +211,12 @@ end
 
     # 228
     @test Tables.columntable(NamedTuple[]) === NamedTuple()
+
+    @testset "Type stability of schema(x).types" begin
+        _get_types(X) = Tables.schema(X).types
+        x = (a=[1.0], b=[1.0])
+        @inferred _get_types(x)
+    end
 end
 
 @testset "Materializer" begin


bkamins:
I have two comments:
1. Have you benchmarked how `Val` scales for wide tables (like 1000+ columns)?
2. Could you please add a test that checks for type stability of the result?

MilesCranmer:
> 1. Have you benchmarked how Val scales for wide tables (like 1000+ columns)?

Keep in mind that this PR does not modify the code's behavior at all. It just switches from `Tuple(... for i=1:n)` to `ntuple(..., Val{n})`, taking advantage of the fact that `n` is a compile-time constant (if `types` is not `nothing`). The return type is the same, it just helps the compiler infer it.

If `types` are nothing, as I guess would happen for a large table like that (?), the `Val` would not even get hit:
```julia
return types === nothing ? (T !== nothing ? T : nothing) : ntuple(i -> fieldtype(types, i), Val(fieldcount(types)))
```

> 2. Could you please add a test that checks for type stability of the result?

Any tips on how to do this? I've never unit-tested compiler type inference before. It feels like something that would be closely tied to a specific Julia version too.

MilesCranmer:
Oh it looks like there is `Test.@inferred`. Is that what you mean?

bkamins:
> The return type is the same, it just helps the compiler infer it.

The point of my question is that I want to make sure that if the compiler tries to infer it it is not too costly. (maybe it does not make a difference, but I want to make sure; in general passing `Val` to `ntuple` usually leads to code generation).

> Oh it looks like there is `Test.@inferred`. Is that what you mean?

Yes

MilesCranmer:
> The point of my question is that I want to make sure that if the compiler tries to infer it it is not too costly. (maybe it does not make a difference, but I want to make sure; in general passing `Val` to `ntuple` usually leads to code generation).

Okay got it. I just tried and there's no difference. I used the `WideTable` unittest to check. Here is the before:
```
julia> @benchmark getproperty(s, :types) setup=(w=WideTable(); s=Tables.schema(w))
BenchmarkTools.Trial: 177 samples with 995 evaluations.
 Range (min … max):  39.112 ns … 42.797 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     39.866 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   39.933 ns ±  0.599 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅▁        ▂▁ █  ▁▆                                           
  ██▆▁▃▄▄▆▅▄██▇█████▆▆█▄▄▄▄▃▄▃▃▄▃▁▃▄▃▃▁▁▄▁▃▁▁▁▃▁▃▁▁▁▁▁▁▁▁▁▁▁▃ ▃
  39.1 ns         Histogram: frequency by time        42.3 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```
and the after
```
julia> @benchmark getproperty(s, :types) setup=(w=WideTable(); s=Tables.schema(w))
BenchmarkTools.Trial: 180 samples with 997 evaluations.
 Range (min … max):  38.406 ns … 60.265 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     39.222 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   39.552 ns ±  1.945 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁   ▆█▇▆▂                                                    
  █▃▃▆█████▆▆▆▃▁▃▃▃▃▃▁▁▃▁▁▁▁▁▃▃▃▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▃
  38.4 ns         Histogram: frequency by time        45.8 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```


> Yes

Done!

MilesCranmer:
[see below; my initial benchmark included compilation time]

MilesCranmer:
Oh, wait, I think that accidentally included compilation time. If you fix the table, it's just fine. It even improves performance:

Before:
```
julia> x = NamedTuple([gensym() => [1.0] for _=1:1000]);

julia> s = Tables.schema(x);

julia> @benchmark getproperty($s, :types)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  15.708 μs … 249.708 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     17.208 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.460 μs ±   3.971 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▁█▆▂                                                   
  ▃▆▄▃▃▂▄████▇▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▁▂▂▂▂ ▃
  15.7 μs         Histogram: frequency by time         25.7 μs <

 Memory estimate: 15.81 KiB, allocs estimate: 2.
```

After:
```
julia> x = NamedTuple([gensym() => [1.0] for _=1:1000]);

julia> s = Tables.schema(x);

julia> @benchmark getproperty($s, :types)
BenchmarkTools.Trial: 10000 samples with 611 evaluations.
 Range (min … max):  197.763 ns … 251.227 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     198.036 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   198.641 ns ±   2.334 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █▃                                                            
  ▇██▆▃▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  198 ns           Histogram: frequency by time          208 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

Edit: forgot the `$s`.

MilesCranmer:
Happened to have some free time this evening so ran some detailed benchmarking; see below. Both TTFX and runtime are shown.

plot

The key takeaways are:
- Runtime:
  - 100-1000x speedup for number of cols < 40
  - 100x speedup for number of cols > 40
- Time to first `.types`:
  - 1-3x speedup for number of cols < 150
  - increasingly slower for number of cols > 150, with 7x slowdown at 1000 columns

So, up to you. I personally think it's worth it, especially because large number of columns are already dealt with in the package.

Furthermore you'd get type stability in downstream packages which will bring even more speedups.

<details><summary>Some of the code used below:</summary>

Benchmarks:
```julia
using Tables
using BenchmarkTools
using Plots

compiled = BenchmarkGroup()
ttfx = Dict{Int,Float64}()

all_n = map(i -> round(Int, 2^i), 1.0:0.5:12.0);

function get_types(X)
    return Tables.schema(X).types
end

# Test run:
test_X = NamedTuple([Symbol("x$(j)") => [1.0] for j = 1:1])
@timed get_types(test_X)

for n in all_n
    local X = NamedTuple([Symbol("x$(j)") => [1.0] for j = 1:n])
    ttfx[n] = (@timed get_types(X)).time
end

for n in all_n
    compiled[n] = @benchmarkable(get_types(X), setup = (X = NamedTuple([Symbol("x$(j)") => [1.0] for j = 1:$n])))
end

tune!(compiled)
compiled_results = run(compiled, verbose=true)

times = [minimum(compiled_results["compiled"][n]).time for n in all_n]

# (Save to CSV file)
```

Plots:
```julia
using Plots
using CSV

# mm for plots:
using Plots.PlotMeasures

ttfx = (
    old=CSV.read("ttfx_old.csv", NamedTuple),
    new=CSV.read("ttfx_new.csv", NamedTuple),
)

ttx = (
    old=CSV.read("ttx_old.csv", NamedTuple),
    new=CSV.read("ttx_new.csv", NamedTuple),
)

theme(:bright)

p = plot(
    plot(
        [ttfx.old.n, ttfx.new.n],
        [ttfx.old.times, ttfx.new.times],
        xaxis=(:log, "Columns in Table"),
        yaxis=(:log, "Time [s]"),
        title="Time to first `.types`",
        label=["old" "new"],
    ),
    plot(
        [ttx.old.n, ttx.new.n],
        [ttx.old.times, ttx.new.times],
        xaxis=(:log, "Columns in Table"),
        yaxis=(:log, "Time [ns]"),
        title="Compiled runtime of `.types`",
        label=["old" "new"],
    ),
    layout=(1,2),
    size=(1000,500),
    # Prevent xlabel being cut off:
    margin=5mm,
    # Darker major grid lines:
    gridalpha=0.5,
)
savefig(p, "plot.png")

```

</details>

bkamins:
> especially because large number of columns are already dealt with in the package.

You mean that if one uses large number of columns then anyway type-unstable table would be used (like `DataFrame`)?

MilesCranmer:
I was just reading that from the docstring:

> Encoding the names & types as type parameters allows convenient use of the type in generated functions and other optimization use-cases, but users should note that when names and/or types are the nothing value, the names and/or types are stored in the storednames and storedtypes fields. This is to account for extremely wide tables with columns in the 10s of thousands where encoding the names/types as type parameters becomes prohibitive to the compiler. So while optimizations can be written on the typed names/types type parameters, users should also consider handling the extremely wide tables by specializing on Tables.Schema{nothing, nothing}.

From which it sounds clear to me that users would not use thousands of columns in the type parameter anyways. (If they did, they would be screwed in other ways as well)

MilesCranmer:
If the compilation time is a big worry for you, I could do this instead:

```julia
n = fieldcount(types)
if n <= 512
    return ntuple(i -> fieldtype(types, i), Val(n))
else
    return Tuple(fieldtype(types, i) for i=1:n)
end
```

since `n` is a compile time constant this would not affect the type stability for less than 512 columns.

MilesCranmer:
I added it. For `ncol>512` it falls back to the type unstable method.

plot

MilesCranmer:
I will be traveling for a bit so please feel free to edit the branch directly and merge if you want

bkamins:
The decision maker is @quinnj here. I wanted to make sure we tested everything before making a decision. I agree that users probably do not use super wide tables in type-stable mode.

MilesCranmer:
Thanks, no worries. In any case with the change in 4d4c9621cdb38d856301a473218b5e3a5a8de974, there is not really any regressions here – in compile time or run time – even for large tables. At worst it's about the same performance; at best you get a 100x-1000x runtime speedup, and downstream packages can do type inference.

MilesCranmer:
Ping @quinnj when you get a chance could you take a look at potentially merging this? Would be a big help for downstream packages

MilesCranmer:
Ping @quinnj

MilesCranmer:
Ping @quinnj 😅

bkamins:
@quinnj is OOO for a few weeks so we need to wait

MilesCranmer:
Gotcha, no worries!

MilesCranmer:
Hey @quinnj, would you be free to take a look at merging this? It’s holding up some type stability issues in MLJ. Thanks, Miles

MilesCranmer:
Sending another ping. This is still a source of type instabilities throughout the MLJ ecosystem, but is a really simple fix.

MilesCranmer:
Ping

MilesCranmer:
Ping

MilesCranmer:
Is there another way to reach @quinnj that he checks more frequently?

bkamins:
I think @quinnj is off-line on all channels currently.

src/Tables.jl
```diff
@@ -499,8 +499,18 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
     if field === :names
         return names === nothing ? getfield(sch, :storednames) : names
     elseif field === :types
-        T = getfield(sch, :storedtypes)
-        return types === nothing ? (T !== nothing ? T : nothing) : Tuple(fieldtype(types, i) for i = 1:fieldcount(types))
+        if types === nothing
+            T = getfield(sch, :storedtypes)
+            return (T !== nothing ? T : nothing)
```
bkamins:
```suggestion
            return T !== nothing ? T : nothing
```

bkamins (approved):

src/Tables.jl
```diff
@@ -499,8 +499,18 @@ function Base.getproperty(sch::Schema{names, types}, field::Symbol) where {names
     if field === :names
         return names === nothing ? getfield(sch, :storednames) : names
     elseif field === :types
-        T = getfield(sch, :storedtypes)
-        return types === nothing ? (T !== nothing ? T : nothing) : Tuple(fieldtype(types, i) for i = 1:fieldcount(types))
+        if types === nothing
+            T = getfield(sch, :storedtypes)
+            return T !== nothing ? T : nothing
```
bkamins: I do not understand the lines in this branch. Why not just `return getfield(sch, :storedtypes)`?

Other than that I can merge this PR.

MilesCranmer: 🤷‍♂️ (I tried to make it as close as possible to the original code to get the PR merged faster)

But yeah let me change it!

MilesCranmer: Done.

MilesCranmer:
Awesome!  Thanks!

bkamins:
do you need a release or `main` branch is enough for now?

MilesCranmer:
A release would be necessary (to get type stability across MLJ ecosystem)

PR #341, opened 2023-08-20
Make it explicit that we require 1-based indexing

bkamins:
In places where we say that some collection is indexable make it explicit that we assume 1-based indexing.

The reason is that downstream packages assume this so I think it is better to document it.


MilesCranmer:
Does it make sense to use `Base.require_one_based_indexing` anywhere here?

bkamins:
@quinnj - OK to merge?

quinnj (approved):

bkamins:
Thank you!

Issue #342, opened 2023-09-06
Add a hint to tell that the row iterator state is the row number in tables with row access


ronisbr:
Hi!

We have a problem in PrettyTables.jl related to huge matrices with row access. The current method, AFAIK, to access an element in those tables is to iterate the rows until we reach the desired one and then retrieve the column:

```julia
    # Get the i-th row by iterating the row table.
    it, state = iterate(rtable.table)

    for _ in 2:i
        it, state = iterate(rtable.table, state)
        it === nothing && error("The row `i` does not exist.")
    end
```

This method is OK for any usage that requires to read the entire data. However, when you are printing huge tables, you usually want a part of the table. In the case of PrettyTables.jl, we have a mode in which we print the beginning and the end of the table. Hence, for each printed column, we need to iterate through all the rows to the very bottom, even though we only need a few of them, leading to a very slow process.

The problem appeared here: https://github.com/ronisbr/PrettyTables.jl/issues/217#issuecomment-1707998660

If we add to the Tables.jl API some hint that the state in the iteration is the row id, I can access, let's say, the element in the row 1,000,000 by just:

```julia
it, ~ = iterate(stable.table, 1_000_000)
```

instead of looping through the entire iterator.

In the case of GeoTables.jl, which uses the row ID as the state, the printing process was reduced from 8s to 0.001s when we only show 10 rows and 10 columns with "middle" cropping mode.

bkamins:
I think there would be a need for "trait" that verifies if a given table has a known number of rows and these rows are indexable (as not all tables support this). E.g. `nothing` would be returned if a table does not support row indexing and otherwise row could be returned.

ablaom:
Yes!

bkamins:
OK. I have looked at it a bit more. So:

1. the `DataAPI.nrow` function is in the API.
2. the `Tables.subset` function is also in the API.

The only question is how well this API is supported in the ecosystem, but I think that the best way forward is to just start using it and then the packages that miss it can add it.

Probably more work needs to be done with `nrow`/`ncol` support which is worse than `Tables.subset` support currently.

(and in PrettyTables.jl maybe what can be done is e.g. try running `nrow` and if it throws `MethodError` do fallback to the slow variant)

bkamins:
[...]

PR #343, opened 2023-09-07
Add nrow and ncol fallback implementations

bkamins:
To help with issues like https://github.com/JuliaData/Tables.jl/issues/342

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -229,19 +229,21 @@ make my custom type valid for Tables.jl consumers?
 
 For a type `MyTable`, the interface to becoming a proper table is straightforward:
 
-| Required Methods                       | Default Definition           | Brief Description                                                                                                               |
-|----------------------------------------|------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
-| `Tables.istable(::Type{MyTable})`      |                              | Declare that your table type implements the interface                                                                           |
-|  **One of:**                           |                              |                                                                                                                                 |
-| `Tables.rowaccess(::Type{MyTable})`    |                              | Declare that your table type defines a `Tables.rows(::MyTable)` method                                                              |
-| `Tables.rows(x::MyTable)`                   |                              | Return an `Tables.AbstractRow`-compatible iterator from your table                                                              |
-| **Or:**                                |                              |                                                                                                                                 |
-| `Tables.columnaccess(::Type{MyTable})` |                              | Declare that your table type defines a `Tables.columns(::MyTable)` method                                                           |
-| `Tables.columns(x::MyTable)`                |                              | Return an `Tables.AbstractColumns`-compatible object from your table                                                            | 
-| **Optional methods**                   |                              |                                                                                                                                 |
-| `Tables.schema(x::MyTable)`                     | `Tables.schema(x) = nothing` | Return a [`Tables.Schema`](@ref) object from your `Tables.AbstractRow` iterator or `Tables.AbstractColumns` object; or `nothing` for unknown schema |
-| `Tables.materializer(::Type{MyTable})` | `Tables.columntable`         | Declare a "materializer" sink function for your table type that can construct an instance of your type from any Tables.jl input |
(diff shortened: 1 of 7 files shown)


bkamins:
I can see that my implementation is incorrect, but let us wait for @quinnj to comment how this should be fixed (my implementation does not work correctly for all tables).

juliohm:
Thank you Bogumil for addressing this issue. It would be really nice to get more explicit support in Tables.jl for `nrow` and `ncol`, especially `nrow`.

A few comments:

1. Julia Base has a different naming convention for functions that return integers: `ndims`, `nfields`, `ndigits`, ... all end with "s". Is there a reason for `nrow` and `ncol` to be singular? Would it be too breaking to rename them to `nrows` and `ncols` for consistency?
2. Is there a reason why `nrow` and `ncol` are implemented in DataAPI.jl? Could these names be moved to Tables.jl in place of `Tables.rowcount`? It feels like the most natural home for these functions.
3. I understand that Tables.jl also targets the use case where tables have infinite streams of rows. I think this use case could be considered the exception instead of the default, and that `nrow` and `ncol` were part of the default table trait. Infinite tables could be formalized with a special trait that only a few users would benefit.
4. It is important to emphasize that adding `nrow` to the default table trait doesn't imply that all tables will fit in memory. We can have the a priori knowledge of `nrow` from a file format specification, and still only be able to load chunks of the file in memory because it is too big.

bkamins:
> Is there a reason for `nrow` and `ncol` to be singular?

This is just a convention that was already in DataFrames.jl before I started developing it. The convention originates from R I guess.
So this is kind of a situation where the names were picked because everyone used them. It is like in the English language (I am not a native speaker so my perspective does not have to be super correct): people say "gonna" because it is short and everyone says so, although I was taught that this is technically incorrect.

> It feels like the most natural home for these functions.

They could. The reason is that Tables.jl provides implementations while DataAPI.jl provides only API. And as such `nrow` and `ncol` do not have currently a default implementation (they could get it in Tables.jl).

> I understand that Tables.jl also targets the use case where tables have infinite streams of rows.

Yes. That is why I have commented above that what should `nrow`/`ncol` return in case of a lazy table that is e.g. schemaless and of unknown number of rows would have to be decided as a part of the design.

> It is important to emphasize that adding `nrow` to the default table trait doesn't imply that all tables will fit in memory.

Yes. `nrow` should play with `Tables.subset` (so that users know what indices are valid)

bkamins:
@quinnj - I have updated the PR following your comment on `rowcount` usage. Can you please have a look if its design is OK now? If yes I will add tests.

docs/src/index.md
```diff
@@ -229,19 +229,21 @@ make my custom type valid for Tables.jl consumers?
...
-| `Tables.columns(x::MyTable)`                |                              | Return an `Tables.AbstractColumns`-compatible object from your table                                                            | 
-| **Optional methods**                   |                              |                                                                                                                                 |
-| `Tables.schema(x::MyTable)`                     | `Tables.schema(x) = nothing` | Return a [`Tables.Schema`](@ref) object from your `Tables.AbstractRow` iterator or `Tables.AbstractColumns` object; or `nothing` for unknown schema |
-| `Tables.materializer(::Type{MyTable})` | `Tables.columntable`         | Declare a "materializer" sink function for your table type that can construct an instance of your type from any Tables.jl input |
-| `Tables.subset(x::MyTable, inds; viewhint)`  |            | Return a row or a sub-table of the original table
+| Required Methods                             | Default Definition           | Brief Description                                                                                                                                   |
```
bkamins, line 232: Here I only added `nrow` and `ncol` lines (at the end). The rest is just the whitespace alignment of a table.

quinnj (approved):
@bkamins, this LGTM; would you mind adding some simple tests that cover the new `nrow`/`ncol` definitions for the various Tables-owned table types? With that, feel free to merge and we can tag a new release.

bkamins:
Thank you!

PR #350, opened 2023-10-20
fix error in error

ericphanson:
tiny fix since `ret` isn't defined. (found by JET!)

### DIFF
--- a/src/dicts.jl
+++ b/src/dicts.jl
@@ -126,14 +126,14 @@ end
 
 function subset(x::DictRowTable, inds; viewhint::Union{Bool,Nothing}=nothing, view::Union{Bool,Nothing}=nothing)
     if view !== nothing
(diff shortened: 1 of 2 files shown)


src/dicts.jl
```diff
@@ -126,14 +126,14 @@ end
 
 function subset(x::DictRowTable, inds; viewhint::Union{Bool,Nothing}=nothing, view::Union{Bool,Nothing}=nothing)
     if view !== nothing
-        @warn "`view` keyword argument is deprecated for `Tables.subset`, use `viewhint` instead"
+        @warn "`view` keyword argument is deprecated for `Tables.subset`, use `viewhint` instead" maxlog=10
```
ericphanson, line 129: unrelated change but deprecation-style warnings should generally get a `maxlog` imo

bkamins (approved):

bkamins:
Thank you!

PR #351, opened 2023-11-08
Add UnROOT to INTEGRATIONS.md replacing UpROOT

Moelf:
uproot is not recommended anymore with UnROOT is much more mature now


bkamins:
I cannot see this recommendation in the readme.md of either of these packages.

Moelf:
the uproot repo has:
>There also exists a pure-Julia ROOT file reading library UnROOT.jl that is built for speed and memory efficiency.

bkamins:
I saw this, but it does not mean that this is preferred. In other words: are you sure that UpRoot.jl maintainers are OK with removing a reference to their package here that we already have?

Moelf:
yes, basically we (JuliaHEP) don't think people should be using Uproot via Python unless for missing issues such as writing out `.root` files, but the Table interface is most useful for reading, and I don't want it to be confusing for new users

bkamins (approved):

PR #355, opened 2024-04-08
Improve handling of vectors of vectors

### LINKED ISSUE #354: materializer behavior differs from its docs
Docs:
```
help?> Tables.materializer
  Tables.materializer(x) => Callable

  <...>
  The default materializer is Tables.columntable, which converts any table input into a NamedTuple of Vectors.

  It is recommended that for users implementing MyType, they define only materializer(::Type{<:MyType}). materializer(::MyType) will then
  automatically delegate to this method.
```
Actual behavior:
```julia
julia> using StructArrays, Tables

julia> sa = StructArray(a=[1,2])


# default: returns columntable for the type, ...
julia> Tables.materializer(StructArray)
columntable (generic function with 5 methods)

# ...  but rowtable for an instance
julia> Tables.materializer(sa)
rowtable (generic function with 1 method)


# if I define materializer() for the type, ...
julia> Tables.materializer(::Type{StructArray}) = StructArray ∘ columntable

# ... it works for the type ...
julia> Tables.materializer(StructArray)
StructArray ∘ Tables.columntable

# ... but still returns rowtable for an instance
julia> Tables.materializer(sa)
rowtable (generic function with 1 method)
```
These are not due to StructArrays doing something weird – it doesn't have `materializer` definitions yet. I noticed these inconsistencies when adding `materializer` support there, https://github.com/JuliaArrays/StructArrays.jl/pull/276.

Would be nice to either fix the behavior to match the docs, or fix the docs. I'm not familiar with Tables internals, so it's hard to tell which is correct.

bkamins:
Fixes https://github.com/JuliaData/Tables.jl/issues/354.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -213,7 +213,8 @@ end
             WT = wrappedtype(eltype(rowitr))
             if WT <: Tuple
                 return allocatecolumns(Schema((Symbol("Column$i") for i = 1:fieldcount(WT)), _fieldtypes(WT)), 0)
-            elseif isconcretetype(WT) && fieldcount(WT) > 0
+            # special case Vector from base and view of this vector as we are sure they do not store rows of a table
+            elseif isconcretetype(WT) && fieldcount(WT) > 0 && !(WT <: Union{Vector, SubArray{<:Vector}})
                 return allocatecolumns(Schema(fieldnames(WT), _fieldtypes(WT)), 0)
             end
         end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -120,6 +120,11 @@ using Test, Tables, OrderedCollections, TableTraits, DataValues, QueryOperators,
     sch = Tables.Schema((), ())
     @test sch.names == ()
     @test sch.types == ()
+
+    # Empty vector of Vector or their view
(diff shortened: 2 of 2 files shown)


bkamins:
Current behavior under Julia nightly:

```
julia> Tables.columntable(Vector{Any}[])
(ref = MemoryRef{Any}[], size = Tuple{Int64}[])

julia> T = typeof(view(Vector{Any}[], :))
SubArray{Vector{Any}, 1, Vector{Vector{Any}}, Tuple{Base.Slice{Base.OneTo{Int64}}}, true}

julia> Tables.columntable(T[])
(parent = Vector{Vector{Any}}[], indices = Tuple{Base.Slice{Base.OneTo{Int64}}}[], offset1 = Int64[], stride1 = Int64[])
```

nalimilan:
Maybe we should apply this to any `AbstractVector`? There's a small risk of breaking something, but it would sound more consistent.

bkamins:
We patched this in DataFrames.jl in https://github.com/JuliaData/DataFrames.jl/pull/3435, so making adjustments here is not critical for DataFrames.jl.

quinnj:
I think this was resolved in #383

PR #356, opened 2024-07-16
remove LinearAlgebra as dependency

longemen3000:
seems that `LinearAlgebra` is loaded but not used. this affects downstream uses (as now LinearAlgebra is loaded separately). related: https://github.com/JuliaFolds2/Transducers.jl/issues/35


quinnj (approved):
Hmmm, weird. I think it was because we used to overload matrix/Matrix, but that didnt' make sense so we just made it `Tables.matrix`? Either way, this is a good clean up. Thanks @longemen3000!

PR #359, opened 2024-11-26
Fix markdown lists in docstring

abhro:
Markdown lists require all items in list to be indented so the list does not break


quinnj (approved):
Thanks!

PR #362, opened 2025-03-05
Improve `show` for `AbtractRow`

omus:
Fixes #361. We could definitely make some improvements beyond this but I think this is overall an improvement on what we had.

PR #363, opened 2025-03-05
Modernize CI workflows

omus:
Noticed CI failures in #362

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -1,0 +1,10 @@
+---
+comment: off
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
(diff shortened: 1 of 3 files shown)


omus:
```
error - 2025-03-05 21:01:59,213 -- Upload failed: {"message":"Token required because branch is protected"}
```
Some work to do for CodeCov

omus:
I have correctly guessed the CodeCov token secret name. Tokenless uploads are possible here but would need to be enabled. See: https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token

omus:
> [Make code coverage informational only](https://github.com/JuliaData/Tables.jl/pull/363/commits/292994c0a3de58a41fc63c6ca94677d4f2538001)

Coverage reductions no longer cause CI checks to fail

quinnj (approved):
thanks for this

PR #364, opened 2025-03-23
Create dependabot.yml

abhro:
From https://discourse.julialang.org/t/psa-use-dependabot-to-update-github-actions-automatically/96001


quinnj (approved):
Thanks!

PR #365, opened 2025-03-24
Documentation updates

abhro:
- Use newer Documenter.jl
- Split up using/implementing Tables.jl into separate doc pages (each were long enough to require their own page)
- Add separate API reference


quinnj (approved):
Thanks @abhro; this is fantastic.

PR #370, opened 2025-05-12
Add doctest in docstring for ByRow()

abhro:


bkamins (approved):
Looks good to me. The changes rely on the Julia version the code is run under, so they might require maintenance in the future.

bkamins:
Thank you

PR #373, opened 2025-08-23
Make AbstractRows print long-ways

pdeffebach:
1. I got rid of printing the type parameters
2. I print each element in a row vertically
3. I do some corrections similar to dataframes about the height and width of the screen. 

An example here [x-ref](https://github.com/JuliaData/DataFrames.jl/issues/2616#issuecomment-3216959296) in DataFrames.jl

```julia
(Tables) julia> using Tables

(Tables) julia> nt = (;(Symbol("x$i") => rand() for i in 1:50)...,);

(Tables) julia> t = Row(nt)
Row: 
x1           0.726270389343219
x2           0.2826873308517256
x3           0.30245552857274705
x4           0.2564103387347969
x5           0.114357002151807
x6           0.7268839730501291
            ⋮
x50          0.43397092645961866
x49          0.40254808328675196
x48          0.934923884837159
x47          0.1640042997187814
x46          0.7799762481712108
x45          0.44932545509234945
x44          0.09937457554982676
```

### DIFF
--- a/src/Tables.jl
+++ b/src/Tables.jl
@@ -201,18 +201,22 @@ function Base.NamedTuple(r::RorC)
 end
 
 function _print_cols(io, inds, nms, max_width, w, h, x)
+    max_ind = maximum(inds)
     for i in inds
         n = nms[i]
         v = x[n]
         s = string(v)
         print(io, rpad(n, max_width + 10, ' '))
         width_renaining = w - 10 - max_width
-        if length(s) < width_renaining
+        if length(s) >= width_renaining
+            s = s[1:(width_renaining)-1] * '⋯'
+        end
+        if i < max_ind
             println(io, s)
         else
-            s_sub = s[1:(width_renaining)-1] * '⋯'
-            println(io, s)
+            print(io, s)
         end
+
     end
 end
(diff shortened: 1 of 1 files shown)


bkamins:
The key decision is if we omit printing data types. I think it is OK, but let us wait for comments.

quinnj:
[Reviewed by AI]

Three regressions reproduced on Julia 1.12.6 at `5fffc3b`, against base `06abc19`:

- Empty rows throw from `maximum`: `sprint(show, Tables.Row(NamedTuple()); context=:limit=>true)` worked on the base. Please handle an empty column list before computing widths/ranges.
- Unicode truncation uses byte indices: `sprint(show, Tables.Row((a=repeat("é", 20),)); context=(:limit=>true, :displaysize=>(24, 14)))` throws `StringIndexError` at `s[1:2]`. Use character-safe truncation and display width.
- The bare `println()` in the truncated-row branch writes to global stdout. Capturing a 30-column row with `show(IOContext(buffer, :limit=>true, :displaysize=>(12, 80)), row)` still prints a newline outside `buffer`. Use `println(io)`.

Please add display tests for these cases. This was a focused display review; I did not run the full package suite.

PR #380, opened 2026-08-15
Add Tables.Scan and generic scan execution

quinnj:
## Summary

Add `Tables.Scan`, a plain-data request for projection, filtering, row bounds, renaming, and output conversion. Add `Tables.scan(table, request)` as the generic reference implementation.

```julia
request = Tables.Scan(
    select = (:id, :price => Float64 => :amount),
    filter = Tables.colcmp(==, Tables.col(:status), "active") &
             (Tables.col(:price) >= 10),
    limit = 100,
)

result = Tables.scan(table, request)
```

A Scan is a logical request. It does not require a physical full-table scan. A source can use indexes, statistics, partition pruning, or other exact optimizations while it reads data. It can then reject unsupported work or pass an unconsumed residual Scan to the generic executor. There is no capability-negotiation, `apply`, or `finish` protocol.

## Consumer semantics

- Requests and expression trees contain plain data. They do not store callbacks.
- The operation order is filter, offset/limit, then projection/rename/type conversion.
- Filters use source column names before rename.
- A filter keeps only rows whose result is exactly `true`; `missing` is excluded, like SQL `WHERE`.
- `select=Tables.All()` is the projection identity and the default.
- `select=()` selects zero columns.
- Selection order defines output order.
- `validate=false` drops unmatched selections and treats unmatched filter columns as all-`missing`.
- An identity request returns the input table unchanged.
- Zero-column results preserve their row count.

Selection supports names, indices, regular expressions, `DataAPI.All()` as `Tables.All()`, `Tables.Not(...)`, output renames, and type overrides.

Filters support `Tables.colcmp`, ordered comparison shorthand, `Tables.colin`, `Tables.isnull`, string predicates, Boolean composition, and source-specific `OpNode` extensions. Comparison and string operation codes use enums.

`Tables.isnull` is deliberate. Adding `Base.ismissing(::Tables.Col)` would specialize Base's broad fallback and invalidate unrelated compiled code when Tables loads. Equality uses `Tables.colcmp(==, ...)`, so expression construction does not change Base equality contracts.

## Source helpers

- `Tables.resolve` resolves and validates a request against source column names.
- The resolved filter normalizes positional references to source names.
- `Tables.filtermask(::BoundScan, table)` can evaluate against predicate-only columns while preserving `validate=false` behavior.
- `Tables.Scan(existing; ...)` builds a residual after a source consumes supported axes.
- `Tables.describe` displays an original request and its residual.
- A source-specific `OpNode` must be consumed before generic resolution.

The manual now separates the consumer API from the source implementation API and documents pushdown, residual, and missing-value semantics.

All APIs remain namespaced under `Tables`. The minimum Julia version is 1.9.

## Review hardening

The review pass also:

- uses `Tables.All()` as the projection identity and `()` as the zero-column selection;
- imports the shared `DataAPI.All` selector;
- replaces the public `isempty(::Scan)` overload with an internal identity check;
- centralizes selector lookup and improves reference errors;
- adds an explicit `@noinline` mask function barrier;
- uses `Tables.allocatecolumn` for output conversion;
- resolves positional filters before predicate-only materialization;
- preserves row counts for empty projections and zero-column sources without row-sized allocation.

## Local validation

- Full Tables.jl suite passes on Julia 1.9 and Julia 1.12.
- The focused Scan suite passes 156 tests on each version.
- Documentation builds successfully.
- `Test.detect_ambiguities(Tables; recursive=true)` reports no ambiguities.
- CSV rewrite Scan tests pass 307 tests on Julia 1.10 and Julia 1.12 after the mechanical API update.
- Arrow rewrite passes 342 facade tests and every acceptance battery on Julia 1.10 and Julia 1.12 after the mechanical API update.
- CI covers minimum, stable, nightly, 32-bit Windows, documentation, and invalidation checks.

Implementation prepared with Claude Code.

Co-authored by Codex

### DIFF
--- a/docs/src/scan.md
+++ b/docs/src/scan.md
@@ -55,7 +55,7 @@ Tables.Scan(select = (
     :id,
     r"^metric_",
     :price => Float64,
-    :timestamp => DateTime => :created_at,
+    :qty => Int => :quantity,
 ))
 ```
 
@@ -122,17 +122,29 @@ bound = Tables.resolve(request, source_names)
 mask = Tables.filtermask(bound, predicate_columns)
 ```
 
-If a source consumes an axis, it removes that axis from the residual. For
-example, a source that performs projection but leaves filtering and row bounds
-for the generic executor can use:
+If a source consumes an axis, it removes that axis from the residual. The
+axes compose in a fixed order — filter, then row bounds, then projection —
+so an axis can only be removed together with every axis that executes before
+it. A source that evaluated the filter itself (for example with
+`Tables.filtermask`) hands the rest to the generic executor:
 
 ```julia
-residual = Tables.Scan(request; select=Tables.All())
-result = Tables.scan(materialized_columns, residual)
+residual = Tables.Scan(request; filter=nothing)
+result = Tables.scan(filtered_columns, residual)
 ```
 
-`Tables.All()` is the projection identity, so the residual does not repeat
-projection work. Only remove work that the source performed exactly.
+A source that also applied `limit`/`offset` to the qualifying rows strips
+those too (`Tables.Scan(request; filter=nothing, limit=nothing, offset=0)`,
+leaving only projection). Two constraints follow from the ordering:
+
+- `limit`/`offset` count qualifying rows, after the filter. A source that
+  cannot consume the filter must leave the row bounds in the residual as
+  well.
+- Projection may be stripped (`select=Tables.All()`, the projection identity)
+  only when no residual filter references a column the projection dropped or
+  renamed — the residual filter still uses source column names.
+
+Only remove work that the source performed exactly.
 
 A statistics check that only prunes impossible partitions does not consume the
 filter.
--- a/docs/src/scan.md
+++ b/docs/src/scan.md
@@ -1,0 +1,157 @@
+# Scan Requests
+
+`Tables.Scan` is a logical request for projection, filtering, row bounds, and
+output conversion. It does not require a physical full-table scan. A source can
+use an index, statistics, partition pruning, or any other exact optimization.
+
+The API has two roles:
+
+- Consumers construct a [`Tables.Scan`](@ref) and pass it to a source or to
+  [`Tables.scan`](@ref).
+- Source implementations inspect, resolve, and consume supported parts of the
+  request while they read data.
+
+## Consumer API
+
+```@example scan
+using Tables
+
+table = (
+    id = [1, 2, 3, 4],
+    status = ["trial", "active", "active", "closed"],
+    price = [5, 20, 12, 30],
+)
+
+request = Tables.Scan(
+    select = (:id, :price => Float64 => :amount),
+    filter = Tables.colcmp(==, Tables.col(:status), "active") &
+             (Tables.col(:price) >= 10),
+    limit = 10,
+)
+
+Tables.scan(table, request)
+```
+
+The operation order is fixed:
+
+1. Resolve references against source column names.
+2. Evaluate the filter.
+3. Apply `offset`, then `limit`, to qualifying rows.
+4. Select, rename, and convert output columns.
+
+Filters always use source names. A rename does not change the name visible to
+the filter.
+
+### Selection
+
+`select=nothing` keeps every column and also means that the projection axis was
+not requested. This distinction matters when a source builds a residual
+request. An explicit `Tables.All()` is a selection item that expands to every
+source column.
+
+A selection accepts names, indices, regular expressions, `Tables.All()`, and
+`Tables.Not(...)`. Pair forms add a type override or output name:
+
+```julia
+Tables.Scan(select = (
+    :id,
+    r"^metric_",
+    :price => Float64,
+    :timestamp => DateTime => :created_at,
+))
+```
+
+Selection order defines output order. Duplicate output names are an error.
+
+### Filter expressions
+
+`Tables.col(ref)` creates a column reference. The supported predicates are:
+
+- `Tables.colcmp(op, column, value)` for `==`, `!=`, `<`, `<=`, `>`, and `>=`.
+- Ordered shorthand such as `Tables.col(:price) >= 10`.
+- `Tables.colin(column, values)` for membership.
+- `Tables.isnull(column)` and its negation for `missing` checks.
+- `startswith`, `endswith`, and `contains` for strings.
+- `&`, `|`, and `!` for Boolean composition.
+
+Expression nodes contain plain data. They do not store callbacks. This makes a
+request inspectable and suitable for serialization, pushdown, and static
+compilation.
+
+### Missing values
+
+Filter evaluation uses SQL-like three-valued logic:
+
+- Comparisons, membership, and string predicates lift a missing column value to
+  `missing`.
+- `Tables.isnull(column)` returns `true` for `missing` and `false` otherwise.
+- `&`, `|`, and `!` propagate `missing` with three-valued Boolean rules.
+- The top-level filter keeps a row only when its result is exactly `true`.
+
+These lifting rules are deliberate. Julia functions other than comparison
+operators do not generally lift `missing` on their own.
+
+The name `Tables.isnull` also avoids defining `Base.ismissing(::Tables.Col)`.
+Such a method would specialize a broad Base fallback and can invalidate
+unrelated compiled code when Tables loads.
+
+### Unmatched references
+
+The default `validate=true` rejects any selection or filter reference that does
+not match the source schema. With `validate=false`, an unmatched selection is
+dropped and an unmatched filter column behaves as an all-missing column. This
+mode supports schema evolution when fields can be absent.
+
+## Source Implementations
+
+A source does not need to support every operation. It can consume the parts it
+can implement exactly and pass the rest to [`Tables.scan`](@ref). It can also
+reject a request that it cannot safely execute.
+
+[`Tables.resolve`](@ref) resolves a request against source names. Its
+`BoundScan` result contains:
+
+- `columns`: selected source indices, output names, and type overrides.
+- `filter`: a filter with positional references normalized to source names.
+- `filtercols`: source indices required by the filter.
+- `limit`, `offset`, and `validate`: the remaining row and validation settings.
+
+The normalized filter can be evaluated over a table containing only the
+`filtercols` columns:
+
+```julia
+bound = Tables.resolve(request, source_names)
+mask = Tables.filtermask(bound, predicate_columns)
+```
+
+If a source consumes an axis, it removes that axis from the residual. For
+example, a source that performs projection but leaves filtering and row bounds
+for the generic executor can use:
+
+```julia
+residual = Tables.Scan(request; select=nothing)
+result = Tables.scan(materialized_columns, residual)
+```
+
+Only remove work that was performed exactly. A statistics check that only
+prunes impossible partitions does not consume the filter.
+
+`Tables.OpNode(name, args)` is the extension point for source-specific,
+plain-data operations. A source can recognize and consume a named operation
+before it calls `Tables.resolve`. Resolution and the generic executor reject an
+unconsumed `OpNode`.
+
+Zero-column results retain their row count. Sources should preserve the same
+property when they return a fully pushed result.
+
+```@docs; canonical = false
+Tables.Scan
+Tables.scan
+Tables.col
+Tables.colcmp
+Tables.colin
+Tables.isnull
+Tables.resolve
+Tables.filtermask
+Tables.describe
+```
--- a/docs/src/scan.md
+++ b/docs/src/scan.md
@@ -44,10 +44,8 @@ the filter.
 
 ### Selection
 
-`select=nothing` keeps every column and also means that the projection axis was
-not requested. This distinction matters when a source builds a residual
-request. An explicit `Tables.All()` is a selection item that expands to every
-source column.
+`select=Tables.All()` is the default and keeps every column. `select=()` selects
+zero columns while preserving the result's row count.
 
 A selection accepts names, indices, regular expressions, `Tables.All()`, and
 `Tables.Not(...)`. Pair forms add a type override or output name:
@@ -129,12 +127,15 @@ example, a source that performs projection but leaves filtering and row bounds
 for the generic executor can use:
 
 ```julia
-residual = Tables.Scan(request; select=nothing)
+residual = Tables.Scan(request; select=Tables.All())
 result = Tables.scan(materialized_columns, residual)
 ```
 
-Only remove work that was performed exactly. A statistics check that only
-prunes impossible partitions does not consume the filter.
+`Tables.All()` is the projection identity, so the residual does not repeat
+projection work. Only remove work that the source performed exactly.
+
+A statistics check that only prunes impossible partitions does not consume the
+filter.
 
 `Tables.OpNode(name, args)` is the extension point for source-specific,
 plain-data operations. A source can recognize and consume a named operation
--- a/src/scan.jl
+++ b/src/scan.jl
@@ -65,7 +65,15 @@ function _selectitem(x::Not)
         throw(ArgumentError("Not references must be Symbol/String/Int/Regex values"))
     return SelectItem(x, nothing, nothing)
 end
-_selectitem(x::Union{Symbol, String, Int, Regex, Not, All}) = SelectItem(x, nothing, nothing)
+function _selectitem(x::All)
+    # DataAPI.All(cols...) with arguments is deprecated DataFrames syntax;
+    # silently selecting every column would be a data-loss trap
+    isempty(x.cols) ||
+        throw(ArgumentError("Tables.All() takes no arguments in a scan selection; " *
+                            "list the columns directly instead of All(cols...)"))
+    return SelectItem(x, nothing, nothing)
+end
+_selectitem(x::Union{Symbol, String, Int, Regex, Not}) = SelectItem(x, nothing, nothing)
 _selectitem(p::Pair{<:Union{Symbol, String, Int, Regex}, Symbol}) = SelectItem(p.first, nothing, p.second)
 _selectitem(p::Pair{<:Union{Symbol, String, Int, Regex}, String}) = SelectItem(p.first, nothing, Symbol(p.second))
 _selectitem(p::Pair{<:Union{Symbol, String, Int, Regex}, <:Type}) = SelectItem(p.first, p.second, nothing)
@@ -581,7 +589,11 @@ filtermask(s::BoundScan, table) = s.filter === nothing ?
 
 _converted(::Nothing, c::AbstractVector) = c
 function _converted(::Type{T}, c::AbstractVector) where {T}
-    eltype(c) <: Union{T, Missing} && return c
+    # eltype Union{} (necessarily empty) is vacuously <: everything, but the
+    # requested output schema still owes the override type
+    if eltype(c) !== Union{} && eltype(c) <: Union{T, Missing}
+        return c
+    end
     anymissing = any(ismissing, c)
     E = anymissing ? Union{T, Missing} : T
     out = allocatecolumn(E, length(c))
--- a/src/scan.jl
+++ b/src/scan.jl
@@ -9,8 +9,9 @@
 # scan=Tables.Scan(...))`, `Arrow.Table(path; scan=...)` — and apply what they
 # can WHILE materializing (skipping columns never parsed, rows never decoded).
 # What a source cannot push down it either rejects with an ArgumentError, or
-# hands to `Tables.scan` as a residual (`Tables.Scan(scan; select=All())`
-# strips the axes it already handled). Sources are free to support only a
+# hands to `Tables.scan` as a residual (e.g. `Tables.Scan(scan; filter=nothing)`
+# after evaluating the filter itself; axes strip in execution order — filter,
+# then row bounds, then projection). Sources are free to support only a
 # subset — the value is the shared vocabulary and the pushdown, not a
 # capability-negotiation protocol.
 #
@@ -283,10 +284,10 @@ qualify, and how many. Plain data all the way down — see `Tables.scan` for
 the generic executor and the module comment for how sources push it down.
 
   * `select`: a column reference or tuple/vector of select items
-    (`ref`, `ref => name`, `ref => Type`, `ref => Type => name`;
-    refs are `Symbol`/`String`/`Int`/`Regex`/`Tables.Not`/`Tables.All`).
-    `Tables.All()` keeps every column; `()` selects zero columns. Selection
-    order = output order.
+    (`ref`, `ref => name`, `ref => Type`, `ref => Type => name`; refs in the
+    pair forms are `Symbol`/`String`/`Int`/`Regex`, while bare `Tables.Not`
+    and `Tables.All()` items stand alone). `Tables.All()` keeps every column;
+    `()` selects zero columns. Selection order = output order.
   * `filter`: an expression built from `Tables.col`; a row is kept iff the
     predicate evaluates to exactly `true` (`missing` excludes, SQL-style).
     Filters see source column names, before renames.
@@ -333,9 +334,16 @@ end
                 limit=scan.limit, offset=scan.offset, validate=scan.validate)
 
 Copy a scan with some axes replaced — how a source builds the RESIDUAL it
-hands to `Tables.scan` after pushing the rest down: a reader that handled
-projection, types and limits itself but cannot filter does
-`Tables.scan(table, Tables.Scan(scan; select=All(), limit=nothing, offset=0))`.
+hands to `Tables.scan` after pushing the rest down. The axes compose in a
+fixed order — filter, then row bounds, then projection — so a source may only
+strip an axis it performed exactly, together with every axis that comes
+before it: a reader that evaluated the filter itself hands over
+`Tables.Scan(scan; filter=nothing)`; one that also applied the row bounds
+hands over `Tables.Scan(scan; filter=nothing, limit=nothing, offset=0)`.
+Projection (`select=All()`) can only be stripped in addition when no residual
+filter references a column the projection dropped or renamed. A reader that
+cannot consume the filter must leave `limit`/`offset` in the residual too —
+they count qualifying rows, not raw rows.
 """
 function Scan(s::Scan; select=s.select, filter=s.filter,
               limit::Union{Nothing, Integer}=s.limit,
--- a/src/scan.jl
+++ b/src/scan.jl
@@ -5,7 +5,7 @@
 #     table, residual = Tables.apply(source, scan)   # source consumes what it can
 #     table′          = Tables.finish(table, residual)  # generic layer does the rest
 #
-# with `Tables.read(source, scan)` composing the two. The default `apply`
+# with `Tables.scan(source, scan)` composing the two. The default `apply`
 # pushes nothing (everything lands in the residual), so every existing
 # Tables.jl source already works — sources override `apply` to earn
 # performance, never correctness.
@@ -203,7 +203,7 @@ _checkpredicate(x) =
 
 A scan request: what to keep, what to call it, how to type it, which rows
 qualify, and how many. Plain data all the way down — see `Tables.apply`,
-`Tables.finish`, and `Tables.read` for the protocol.
+`Tables.finish`, and `Tables.scan` for the protocol.
 
   * `select`: a column reference or tuple/vector of select items
     (`ref`, `ref => name`, `ref => Type`, `ref => Type => name`;
@@ -519,12 +519,15 @@ nothing, so every Tables.jl source works unmodified. Contract:
 apply(source, scan::Scan) = (source, scan)
 
 """
-    Tables.read(source, scan::Scan)
+    Tables.scan(source, scan::Scan)
 
 `Tables.finish(Tables.apply(source, scan)...)` — the one-call entry point.
+Sources that push scans down (a CSV reader, a database) implement
+`Tables.apply`; `Tables.scan` composes it with the generic `finish` so
+callers never see the residual.
 """
-function read(source, scan::Scan)
-    t, residual = apply(source, scan)
+function scan(source, s::Scan)
+    t, residual = apply(source, s)
     return finish(t, residual)
 end
 
--- a/src/scan.jl
+++ b/src/scan.jl
@@ -18,8 +18,8 @@
 #     prunes by a predicate but cannot guarantee exactness keeps the filter in
 #     the residual).
 #   * Missing follows SQL WHERE: a predicate evaluating to `missing` excludes
-#     the row. `isnull`/`!isnull` are first-class nodes; `==` never matches
-#     missing.
+#     the row. `ismissing`/`!ismissing` are first-class nodes; `coleq` never
+#     matches missing.
 #   * Filters reference SOURCE column names (pre-rename); the pipeline order
 #     is fixed: bind → filter → offset/limit → project/rename/type.
 #   * Selection order defines output order.
@@ -62,6 +62,12 @@ struct SelectItem
     rename::Union{Nothing, Symbol}
 end
 
+function _selectitem(x::Not)
+    refs = x.ref isa Union{Tuple, AbstractVector} ? x.ref : (x.ref,)
+    all(r -> r isa ColRef, refs) ||
+        throw(ArgumentError("Not references must be Symbol/String/Int/Regex values"))
+    return SelectItem(x, nothing, nothing)
+end
 _selectitem(x::Union{Symbol, String, Int, Regex, Not, All}) = SelectItem(x, nothing, nothing)
 _selectitem(p::Pair{<:Union{Symbol, String, Int, Regex}, Symbol}) = SelectItem(p.first, nothing, p.second)
 _selectitem(p::Pair{<:Union{Symbol, String, Int, Regex}, String}) = SelectItem(p.first, nothing, Symbol(p.second))
@@ -83,7 +89,7 @@ abstract type ScanExpr end
     Tables.col(ref)
 
 A column reference inside a `Scan` filter: `col(:price) > 100`. Comparisons
-against literals, `in_`, `isnull`, `startswith`/`endswith`/`contains`, and
+against literals, `in_`, `ismissing`, `startswith`/`endswith`/`contains`, and
 `&`/`|`/`!` build plain expression values.
 """
 struct Col <: ScanExpr
@@ -98,7 +104,32 @@ struct Cmp{T} <: ScanExpr
     op::UInt8
     lhs::Col
     rhs::T
+    function Cmp(op::Integer, lhs::Col, rhs::T) where {T}
+        OP_EQ <= op <= OP_GE ||
+            throw(ArgumentError("unknown comparison operator code $op"))
+        return new{T}(UInt8(op), lhs, rhs)
+    end
 end
+
+"""
+    Tables.coleq(col, value)
+    Tables.colne(col, value)
+    Tables.collt(col, value)
+    Tables.colle(col, value)
+    Tables.colgt(col, value)
+    Tables.colge(col, value)
+
+Build a comparison predicate for a column and a literal value. These named
+constructors accept any literal type. `<`, `<=`, `>`, and `>=` provide shorthand
+for numeric, string, and character values. Equality uses `coleq` and `colne` so
+`Base.isequal` keeps its Boolean contract.
+"""
+coleq(c::Col, value) = Cmp(OP_EQ, c, value)
+colne(c::Col, value) = Cmp(OP_NE, c, value)
+collt(c::Col, value) = Cmp(OP_LT, c, value)
+colle(c::Col, value) = Cmp(OP_LE, c, value)
+colgt(c::Col, value) = Cmp(OP_GT, c, value)
+colge(c::Col, value) = Cmp(OP_GE, c, value)
 
 struct In{T} <: ScanExpr
     lhs::Col
@@ -112,18 +143,21 @@ Membership predicate: `in_(col(:status), ("active", "trial"))`.
 """
 in_(c::Col, values) = In(c, values)
 
-struct IsNull <: ScanExpr
+struct IsMissing <: ScanExpr
     lhs::Col
     negated::Bool
 end
 
 """
-    Tables.isnull(col)
+    Tables.ismissing(col)
 
-Missing-ness predicate. Use this — never `col(:x) == missing`, which follows
-SQL semantics and matches no row.
+Missing-ness predicate. Use this rather than `coleq(col(:x), missing)`, which
+follows SQL semantics and matches no row.
 """
-isnull(c::Col) = IsNull(c, false)
+# Keep this function Tables-local. Base.ismissing has an Any fallback, so a Col
+# method there invalidates unrelated precompiled code.
+function ismissing end
+ismissing(c::Col) = IsMissing(c, false)
 
 const STR_STARTSWITH, STR_ENDSWITH, STR_CONTAINS = 0x01, 0x02, 0x03
 
@@ -131,6 +165,11 @@ struct StrPred <: ScanExpr
     kind::UInt8
     lhs::Col
     s::String
+    function StrPred(kind::Integer, lhs::Col, s::AbstractString)
+        STR_STARTSWITH <= kind <= STR_CONTAINS ||
+            throw(ArgumentError("unknown string predicate code $kind"))
+        return new(UInt8(kind), lhs, String(s))
+    end
 end
 
 struct AndExpr <: ScanExpr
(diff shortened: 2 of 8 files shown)


src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+#     the row. `isnull`/`!isnull` are first-class nodes; `coleq` never
+#     matches missing.
+#   * Filters reference SOURCE column names (pre-rename); the pipeline order
+#     is fixed: bind → filter → offset/limit → project/rename/type.
+#   * Selection order defines output order.
+
+# --- column references & selection items --------------------------------------
+
+"""
+    Tables.All()
```
nalimilan: Import `DataAPI.All` and similar?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+col(r::Union{Symbol, AbstractString, Int}) = Col(r isa AbstractString ? String(r) : r)
+
+const OP_EQ, OP_NE, OP_LT, OP_LE, OP_GT, OP_GE = 0x01, 0x02, 0x03, 0x04, 0x05, 0x06
+const _OPNAMES = ("==", "!=", "<", "<=", ">", ">=")
+
+struct Cmp{T} <: ScanExpr
+    op::UInt8
+    lhs::Col
+    rhs::T
+    function Cmp(op::Integer, lhs::Col, rhs::T) where {T}
```
nalimilan: Use an enum instead of `Integer`?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+        OP_EQ <= op <= OP_GE ||
+            throw(ArgumentError("unknown comparison operator code $op"))
+        return new{T}(UInt8(op), lhs, rhs)
+    end
+end
+
+"""
+    Tables.coleq(col, value)
```
nalimilan: What's the point of defining all these functions? If anything, I'd rather have a more intuitive API like `Tables.colcmp(==, col, value)` and so on. That would also simplify the implementation below.

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+    Tables.colne(col, value)
+    Tables.collt(col, value)
+    Tables.colle(col, value)
+    Tables.colgt(col, value)
+    Tables.colge(col, value)
+
+Build a comparison predicate for a column and a literal value. These named
+constructors accept any literal type. `<`, `<=`, `>`, and `>=` provide shorthand
+for numeric, string, and character values. Equality uses `coleq` and `colne` so
+`Base.isequal` keeps its Boolean contract.
```
nalimilan: I don't understand these explanations, it looks like it refers to a context that one doesn't have when just reading this docstring.

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+
+struct In{T} <: ScanExpr
+    lhs::Col
+    values::T
+end
+
+"""
+    Tables.in_(col, values)
+
+Membership predicate: `in_(col(:status), ("active", "trial"))`.
```
nalimilan: The name ending with an underscore is a bit weird. Better either use a common prefix for all predicates including `in`, or put them in a submodule with a short name?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+end
+
+"""
+    Tables.isnull(col)
+
+Build a predicate that is true when the column value is Julia's `missing`.
+Use `!Tables.isnull(col)` to match values that are not `missing`.
+
+The `isnull` name is deliberate. Defining `Base.ismissing(::Col)` would
+specialize Base's broad fallback and invalidate unrelated precompiled code
```
nalimilan, line 182: Likewise, I'd rather keep the `ismissing` terminology for clarity, and add a prefix to that term or use a submodule.

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+    Tables.Scan(; select=nothing, filter=nothing, limit=nothing, offset=nothing, validate=true)
+
+A scan request: what to keep, what to call it, how to type it, which rows
+qualify, and how many. Plain data all the way down — see `Tables.apply`,
+`Tables.finish`, and `Tables.scan` for the protocol.
+
+  * `select`: a column reference or tuple/vector of select items
+    (`ref`, `ref => name`, `ref => Type`, `ref => Type => name`;
+    refs are `Symbol`/`String`/`Int`/`Regex`/`Tables.Not`/`Tables.All`).
+    `nothing` keeps every column. Selection order = output order.
```
nalimilan: Use `All()` by default rather than `nothing`?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+end
+
+"""
+    isempty(scan::Tables.Scan)
+
+`true` when the scan requests nothing: no selection, no filter, no limit or
+offset. `Tables.finish` returns the table unchanged for an empty residual.
+"""
+Base.isempty(s::Scan) =
+    s.select === nothing && s.filter === nothing && s.limit === nothing && s.offset == 0
```
nalimilan: This definition doesn't seem super intuitive to me. Naively, I would expect `isempty` to say whether the result contains any rows (like a collection). This seems like `isidentity` or `isnoop`. Better have a separate function? Maybe it could even remain internal?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+    arg::ScanExpr
+end
+struct AlwaysTrue <: ScanExpr end
+struct AlwaysFalse <: ScanExpr end
+
+"""
+    Tables.OpNode(name, args)
+
+The expression algebra's growth channel: a named node with child values.
+Sources that recognize `name` may push it down; the generic executor rejects
```
nalimilan: Why "may"? How can source avoid passing a `name` given that it's `::Symbol`?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+    @eval begin
+        Base.:<(c::Col, v::$T) = collt(c, v)
+        Base.:<(v::$T, c::Col) = colgt(c, v)
+        Base.:<=(c::Col, v::$T) = colle(c, v)
+        Base.:<=(v::$T, c::Col) = colge(c, v)
+    end
+end
+Base.:(==)(a::Col, b::Col) = a.ref == b.ref
+Base.isequal(a::Col, b::Col) = isequal(a.ref, b.ref)
+Base.hash(c::Col, h::UInt) = hash(c.ref, hash(:TablesCol, h))
```
nalimilan: Usually we hash a random hardcoded value instead of a symbol I think?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+Base.:!(e::NotExpr) = e.arg
+Base.:!(e::IsNull) = IsNull(e.lhs, !e.negated)
+
+# a bare Col is not a predicate; catch the likely mistake early
+_checkpredicate(::Nothing) = nothing
+_checkpredicate(e::ScanExpr) = e
+_checkpredicate(e::Col) =
+    throw(ArgumentError("a bare column reference is not a predicate; compare it against a value"))
+_checkpredicate(x) =
+    throw(ArgumentError("filter must be a Tables.Scan expression (built from Tables.col), got $(typeof(x))"))
```
nalimilan:
```suggestion
    throw(ArgumentError("filter must be a Tables.Col expression (built from Tables.col), got $(typeof(x))"))
```

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+        _filterrefs!(refs, e.arg, names, validate)
+    elseif e isa OpNode
+        throw(ArgumentError("cannot bind extension node $(repr(e.name)): no generic evaluation; " *
+                            "only sources that recognize it may consume it"))
+    end
+    return refs
+end
+
+"""
+    Tables.bind(scan::Scan, names) -> BoundScan
```
nalimilan: Could we find a more explicit term than "bind"? In dplyr, `bind_rows` and `bind_cols` concatenate tables (like R's `rbind` and `cbind`). Maybe "resolve"? StatsModels has `apply_schema`, but that also uses information on the types of columns so it's a bit different.

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+            if r isa Regex
+                found = false
+                for (i, nm) in enumerate(nms)
+                    if occursin(r, String(nm))
+                        push!(excluded, i)
+                        found = true
+                    end
+                end
+                scan.validate && !found &&
+                    throw(ArgumentError("Not pattern $(_refstr(r)) matches no column"))
```
nalimilan: To avoid misunderstandings:
```suggestion
                    throw(ArgumentError("Not(...) pattern $(_refstr(r)) matches no column"))
```

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+            else
+                i = _findcol(nms, r)
+                i === nothing && scan.validate &&
+                    throw(ArgumentError("Not reference $(_refstr(r)) matches no column"))
```
nalimilan:
```suggestion
                    throw(ArgumentError("Not(...) reference $(_refstr(r)) matches no column"))
```

src/scan.jl
nalimilan: Why not make this block a `_findcols` method?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+
+_getcol(cols, ref::Symbol) = getcolumn(cols, ref)
+_getcol(cols, ref::String) = getcolumn(cols, Symbol(ref))
+_getcol(cols, ref::Int) = getcolumn(cols, ref)
+
+# Resolve a filter leaf's column. Under `strict` an unknown reference is an
+# error; otherwise it resolves to `nothing` and the leaf evaluates as an
+# ALL-MISSING column — the schema-evolution semantics: a column this table
+# does not have reads as missing everywhere (`isnull` keeps every row, every
+# comparison excludes, and three-valued `&`/`|` compose from there).
```
nalimilan: I don't understand the parenthesis.

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+               throw(ArgumentError("unknown comparison operator code $(e.op)"))
+    elseif e isa In
+        a = _resolvecol(cols, e.lhs.ref, strict)
+        a === nothing && return fill(missing, rowcount(cols))
+        return in.(a, Ref(e.values))
+    elseif e isa IsNull
+        a = _resolvecol(cols, e.lhs.ref, strict)
+        a === nothing &&
+            return e.negated ? falses(rowcount(cols)) : trues(rowcount(cols))
+        return e.negated ? .!Base.ismissing.(a) : Base.ismissing.(a)
```
nalimilan: Why the Base prefix here and elsewhere?

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+    elseif e isa StrPred
+        a = _resolvecol(cols, e.lhs.ref, strict)
+        a === nothing && return fill(missing, rowcount(cols))
+        f = e.kind == STR_STARTSWITH ? startswith :
+            e.kind == STR_ENDSWITH ? endswith :
+            e.kind == STR_CONTAINS ? contains :
+            throw(ArgumentError("unknown string predicate code $(e.kind)"))
+        return map(x -> Base.ismissing(x) ? missing : f(x, e.s), a)
```
nalimilan: These lifting semantics should be documented very clearly as they are an exception to how `missing` behaves in plain Julia (equality operators lift by default but other functions don't).

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+               e.op == OP_NE ? (a .!= v) :
+               e.op == OP_LT ? (a .< v) :
+               e.op == OP_LE ? (a .<= v) :
+               e.op == OP_GT ? (a .> v) :
+               e.op == OP_GE ? (a .>= v) :
+               throw(ArgumentError("unknown comparison operator code $(e.op)"))
+    elseif e isa In
+        a = _resolvecol(cols, e.lhs.ref, strict)
+        a === nothing && return fill(missing, rowcount(cols))
+        return in.(a, Ref(e.values))
```
nalimilan: Note that this differs from SQL semantics where `NULL IN (...)` [gives `NULL`](https://www.postgresql.org/docs/current/functions-comparisons.html#FUNCTIONS-COMPARISONS-IN-SCALAR). It would make sense to follow SQL, though it differs from what Julia does.

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+"""
+# The Bool-mask comprehension runs behind a FUNCTION BARRIER: `_evalexpr`
+# returns an abstractly-typed vector (its result type depends on the
+# expression tree), and `Bool[x === true for x in v]` over an abstract `v`
+# melts into per-element dynamic dispatch — measured 100× slower than the
+# same loop behind a barrier (62 ms → 0.6 ms over a 1M-row mask). The
+# barrier costs ONE dynamic dispatch per mask instead. The vectorized
+# kernels inside `_evalexpr` need no such treatment: broadcast and `map`
+# already specialize on the concrete container at their own call boundary.
+_boolmask(v::AbstractVector) = Bool[x === true for x in v]
```
nalimilan: Maybe safer to do this to make sure inlining doesn't kill performance in the future?
```suggestion
@noinline _boolmask(v::AbstractVector) = Bool[x === true for x in v]
```

src/scan.jl
```diff
@@ -0,0 +1,648 @@
...
+filtermask(s::Scan, table) = s.filter === nothing ?
+    trues(rowcount(columns(table))) :
+    _boolmask(_evalexpr(s.filter, columns(table), s.validate))
+
+_converted(::Nothing, c::AbstractVector) = c
+function _converted(::Type{T}, c::AbstractVector) where {T}
+    eltype(c) <: Union{T, Missing} && return c
+    anymissing = any(Base.ismissing, c)
+    E = anymissing ? Union{T, Missing} : T
+    return E[Base.ismissing(x) ? missing : convert(T, x) for x in c]
```
nalimilan:
```suggestion
    res = allocatecolumn(E, length(c))
    res .= ifelse.(ismissing.(c), missing, convert.(Ref(T), x))
    return res
```

src/scan.jl
```diff
@@ -0,0 +1,659 @@
...
+
+  * `select`: a column reference or tuple/vector of select items
+    (`ref`, `ref => name`, `ref => Type`, `ref => Type => name`;
+    refs are `Symbol`/`String`/`Int`/`Regex`/`Tables.Not`/`Tables.All`).
+    `nothing` keeps every column. Selection order = output order.
+  * `filter`: an expression built from `Tables.col`; a row is kept iff the
+    predicate evaluates to exactly `true` (`missing` excludes, SQL-style).
+    Filters see source column names, before renames.
+  * `limit`/`offset`: applied to qualifying rows, after the filter.
+  * `validate`: error on select/filter references that match no column.
```
nalimilan, line 295: `validate` isn't the most explicit term for this. But unfortunately I haven't been able to find a better one... `allow_unmatched`?

nalimilan (commented):
Interesting!

Here's my review. A few main remarks:
- Have you considered putting this in TableOperations.jl instead? Can this be used in interaction with functions from that package?
- Could this be extended later to allow applying custom functions to any element? This sounds like a major useful feature, even if all backends cannot support it (or not immediately/not natively).
- The name "scan" doesn't seem great to me. It's not super evocative outside of SQL, and in SQL ["full table scan"](https://en.wikipedia.org/wiki/Full_table_scan) refers to an operation which is generally slow as it goes over all rows without using an index. AFAICT a SQL-backed Tables.jl source could use an index under the hood e.g. if one wants to filter rows based on an index column, right? Also if you just use `Scan` to select and rename columns, no iteration over rows happens.
- Docstrings could be improved, and the manual should present the mechanism in general. It's important to distinguish what is supposed to be used by consumers and what is only intended at implementations.

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+#   * Missing follows SQL WHERE: a predicate evaluating to `missing` excludes
+#     the row. `isnull`/`!isnull` are first-class nodes; comparisons never
+#     match missing.
+#   * Filters reference SOURCE column names (pre-rename); the pipeline order
+#     is fixed: resolve → filter → offset/limit → project/rename/type.
+#   * Selection order defines output order.
+
+# --- column references & selection items --------------------------------------
+
+import DataAPI: All
```
nalimilan, line 34: Put imports in src/Tables.jl?

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+table = (
+    id = [1, 2, 3, 4],
+    status = ["trial", "active", "active", "closed"],
+    price = [5, 20, 12, 30],
+)
+
+request = Tables.Scan(
+    select = (:id, :price => Float64 => :amount),
+    filter = Tables.colcmp(==, Tables.col(:status), "active") &
+             (Tables.col(:price) >= 10),
```
nalimilan, line 28: Looking at this example, it's weird that `>=` can be used directly but `==` has to go though `colcmp`. I guess you didn't implement `==` as it would give weird results in many situations? Probably better not implement `>=` and similar then for consistency.

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+
+Filters always use source names. A rename does not change the name visible to
+the filter.
+
+### Selection
+
+`select=nothing` keeps every column and also means that the projection axis was
+not requested. This distinction matters when a source builds a residual
+request. An explicit `Tables.All()` is a selection item that expands to every
+source column.
```
nalimilan: This can't be understood at this point as residual request hasn't been mentioned. Anyway that's an implementation detail for consumers, right?

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+)
+
+Tables.scan(table, request)
+```
+
+The operation order is fixed:
+
+1. Resolve references against source column names.
+2. Evaluate the filter.
+3. Apply `offset`, then `limit`, to qualifying rows.
```
nalimilan, line 39: `offset` and `limit` haven't been mentioned yet. Maybe briefly explain at the top of the section what are the different arguments/operations?

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+Selection order defines output order. Duplicate output names are an error.
+
+### Filter expressions
+
+`Tables.col(ref)` creates a column reference. The supported predicates are:
+
+- `Tables.colcmp(op, column, value)` for `==`, `!=`, `<`, `<=`, `>`, and `>=`.
+- Ordered shorthand such as `Tables.col(:price) >= 10`.
+- `Tables.colin(column, values)` for membership.
+- `Tables.isnull(column)` and its negation for `missing` checks.
```
nalimilan, line 71:
```suggestion
- `Tables.isnull(column)` and its negation to check for `missing`.
```

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+Filter evaluation uses SQL-like three-valued logic:
+
+- Comparisons, membership, and string predicates lift a missing column value to
+  `missing`.
+- `Tables.isnull(column)` returns `true` for `missing` and `false` otherwise.
+- `&`, `|`, and `!` propagate `missing` with three-valued Boolean rules.
+- The top-level filter keeps a row only when its result is exactly `true`.
+
+These lifting rules are deliberate. Julia functions other than comparison
+operators do not generally lift `missing` on their own.
```
nalimilan, line 90: Mathematical operators also lift, so the deviation isn't so great:
```suggestion
These lifting rules are the same as those followed by Julia for operators,
but they differ from the behavior of Julia membership and string predicates,
which do not lift `missing` on their own.
```

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+
+The name `Tables.isnull` also avoids defining `Base.ismissing(::Tables.Col)`.
+Such a method would specialize a broad Base fallback and can invalidate
+unrelated compiled code when Tables loads.
+
+### Unmatched references
+
+The default `validate=true` rejects any selection or filter reference that does
+not match the source schema. With `validate=false`, an unmatched selection is
```
nalimilan, line 99:
```suggestion
The default `validate=true` rejects any selection of or filter reference to a column that
does not exist in the source schema. With `validate=false`, an unmatched selection is
```

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
+# Scan Requests
```
nalimilan, line 1: Given how the docs are currently organized, it seems it would make more sense to split this into a part that would go under "Using the Interface" and another one that would go under "Implementing the Interface"?

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+A source does not need to support every operation. It can consume the parts it
+can implement exactly and pass the rest to [`Tables.scan`](@ref). It can also
+reject a request that it cannot safely execute.
+
+[`Tables.resolve`](@ref) resolves a request against source names. Its
+`BoundScan` result contains:
+
+- `columns`: selected source indices, output names, and type overrides.
+- `filter`: a filter with positional references normalized to source names.
+- `filtercols`: source indices required by the filter.
```
nalimilan, line 114:
```suggestion
- `filtercols`: source column indices required by the filter.
```

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+
+The normalized filter can be evaluated over a table containing only the
+`filtercols` columns:
+
+```julia
+bound = Tables.resolve(request, source_names)
+mask = Tables.filtermask(bound, predicate_columns)
+```
+
+If a source consumes an axis, it removes that axis from the residual. For
```
nalimilan: "Axis" hasn't been defined above.

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
+# Scan Requests
+
+`Tables.Scan` is a logical request for projection, filtering, row bounds, and
```
nalimilan, line 3: "Projection" isn't a term frequently used in Julia, maybe explain that it means selecting/renaming columns? Also "row bounds" isn't clear to me, and "output conversion" could be "output *type* conversion".

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+mask = Tables.filtermask(bound, predicate_columns)
+```
+
+If a source consumes an axis, it removes that axis from the residual. For
+example, a source that performs projection but leaves filtering and row bounds
+for the generic executor can use:
+
+```julia
+residual = Tables.Scan(request; select=nothing)
+result = Tables.scan(materialized_columns, residual)
```
nalimilan:
```suggestion
materialized_columns = ... # select columns
result = Tables.scan(materialized_columns, residual)
```

docs/src/scan.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+Only remove work that was performed exactly. A statistics check that only
+prunes impossible partitions does not consume the filter.
+
+`Tables.OpNode(name, args)` is the extension point for source-specific,
+plain-data operations. A source can recognize and consume a named operation
```
nalimilan, line 153: I don't really understand what this means.

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+
+"""
+    Tables.Not(ref)
```
nalimilan, line 37: Use `InvertedIndices.Not`? This is what DataFrames does (the package is very small).

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+struct Not{T}
+    ref::T
+end
+
+const ColRef = Union{Symbol, String, Int, Regex}
+
+"""
+    Tables.SelectItem
+
+One resolved element of `Scan.select`: a column reference plus optional type
```
nalimilan, line 52:
```suggestion
One resolved element of [`Scan.select`](@ref): a column reference plus optional type
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+override and output name. Users never construct these directly — the `Scan`
```
nalimilan, line 53:
```suggestion
override and output name. Users never construct these directly — the [`Scan`](@ref)
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+
+# --- the expression algebra ----------------------------------------------------
+#
+# A small closed vocabulary — the intersection every surveyed engine supports
+# and every statistics pruner can use. Growth happens through `OpNode`
+# (name + children), never through new struct kinds every consumer must learn.
+
+abstract type ScanExpr end
+
+struct Col <: ScanExpr
```
nalimilan, line 94: Now I realize that DataAPI has `Cols` for the same purpose (plural because a regular expression can select multiple columns)`, which is used by DataFrames. Reuse the same? Cc: @bkamins

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+    ref::Union{Symbol, String, Int}
+end
+"""
```
nalimilan, line 97:
```suggestion

"""
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+    Tables.col(ref)
+
+A column reference inside a `Scan` filter: `col(:price) > 100`. Comparisons
```
nalimilan, line 100:
```suggestion
A column reference inside a [`Scan`](@ref) filter: for example
`col(:price) > 100`. Comparisons
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+    rhs::T
+    function Cmp(op::ComparisonOperator, lhs::Col, rhs::T) where {T}
+        rhs isa Col && _colcolerr()
+        return new{T}(op, lhs, rhs)
+    end
+end
+function Cmp(op::Integer, lhs::Col, rhs)
+    0x01 <= op <= 0x06 || throw(ArgumentError("unknown comparison operator code $op"))
+    return Cmp(ComparisonOperator(op), lhs, rhs)
+end
```
nalimilan, line 130: Is this method really needed?

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+_comparisonoperator(::typeof(==)) = OP_EQ
+_comparisonoperator(::typeof(!=)) = OP_NE
+_comparisonoperator(::typeof(<)) = OP_LT
+_comparisonoperator(::typeof(<=)) = OP_LE
+_comparisonoperator(::typeof(>)) = OP_GT
+_comparisonoperator(::typeof(>=)) = OP_GE
+_comparisonoperator(op) = throw(ArgumentError(
+    "unsupported comparison function $(repr(op)); expected ==, !=, <, <=, >, or >=",
+))
+"""
```
nalimilan, line 141:
```suggestion

"""
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+    Tables.colcmp(op, col, value)
+
+Build a comparison predicate for a column and a literal value. `op` must be
+`==`, `!=`, `<`, `<=`, `>`, or `>=`. Ordered comparisons also support the
+shorthand `col(:x) < value` for numeric, string, and character values.
+
+Use `colcmp` for equality so `==` and `isequal` on expression objects keep
+their normal Boolean contracts.
```
nalimilan, line 149:
```suggestion
`==`, `!=`, `<`, `<=`, `>`, or `>=`.

Ordered comparisons can also be performed directly using operators
(e.g. `col(:x) < value`) for numeric, string, and character values.
But `colcmp` has to be used for equality, as `==` and `isequal` on
expression objects return a Boolean rather than an expression.
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+    function In(lhs::Col, values::T) where {T}
+        values isa Col && _colcolerr()
+        return new{T}(lhs, values)
+    end
+end
+
+"""
+    Tables.colin(col, values)
+
+Build a membership predicate: `colin(col(:status), ("active", "trial"))`.
```
nalimilan, line 165:
```suggestion
Build a membership predicate to be used inside a [`Scan`](@ref) filter:
for example `colin(col(:status), ("active", "trial"))`.
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+If the column value is `missing`, the predicate result is also `missing`.
```
nalimilan, line 166:
```suggestion
If the column value is `missing`, the predicate result is also `missing`
(this differs from Julia's `in`).
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+
+struct IsNull <: ScanExpr
+    lhs::Col
+    negated::Bool
+end
+
+"""
+    Tables.isnull(col)
+
+Build a predicate that is true when the column value is Julia's `missing`.
```
nalimilan, line 178:
```suggestion
Build a predicate to test for missing values inside a [`Scan`](@ref) filter.
The result is true when the column value is `missing` and false otherwise.
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+
+struct Col <: ScanExpr
+    ref::Union{Symbol, String, Int}
+end
+"""
+    Tables.col(ref)
+
+A column reference inside a `Scan` filter: `col(:price) > 100`. Comparisons
+against literals, `colin`, `isnull`, `startswith`/`endswith`/`contains`, and
+`&`/`|`/`!` build plain expression values.
```
nalimilan, line 102:
```suggestion
`&`/`|`/`!` build plain expression values when called on a `col` reference.
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+_comparisonoperator(::typeof(<=)) = OP_LE
+_comparisonoperator(::typeof(>)) = OP_GT
+_comparisonoperator(::typeof(>=)) = OP_GE
+_comparisonoperator(op) = throw(ArgumentError(
+    "unsupported comparison function $(repr(op)); expected ==, !=, <, <=, >, or >=",
+))
+"""
+    Tables.colcmp(op, col, value)
+
+Build a comparison predicate for a column and a literal value. `op` must be
```
nalimilan, line 144:
```suggestion
Build a comparison predicate for a column and a literal value
to be used inside a [`Scan`](@ref) filter. `op` must be
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+    lhs::Col
+    s::String
+    function StrPred(kind::StringPredicate, lhs::Col, s::AbstractString)
+        return new(kind, lhs, String(s))
+    end
+end
+function StrPred(kind::Integer, lhs::Col, s::AbstractString)
+    0x01 <= kind <= 0x03 || throw(ArgumentError("unknown string predicate code $kind"))
+    return StrPred(StringPredicate(kind), lhs, s)
+end
```
nalimilan, line 206: Needed?

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+struct NotExpr <: ScanExpr
+    arg::ScanExpr
+end
+struct AlwaysTrue <: ScanExpr end
+struct AlwaysFalse <: ScanExpr end
+
+"""
+    Tables.OpNode(name, args)
+
+The expression algebra's growth channel. Every node has a symbolic `name` and
```
nalimilan, line 223: Not super clear.

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+end
+
+_isidentity(s::Scan) =
+    s.select === nothing && s.filter === nothing && s.limit === nothing && s.offset == 0
+
+
+# --- schema resolution -----------------------------------------------------------
+
+"""
+    Tables.BoundColumn
```
nalimilan, line 370: Call this `ResolvedColumn` and ResolvedScan` now that the function is `resolve`?

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+struct BoundColumn
+    index::Int
+    name::Symbol
+    type::Union{Nothing, Type}
+end
+
+"""
+    Tables.BoundScan
+
+A `Scan` resolved against a concrete schema: output columns in order, the
```
nalimilan, line 384:
```suggestion
A [`Scan`](@ref) resolved against a concrete schema: output columns in order, the
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+        throw(ArgumentError("cannot resolve extension node $(repr(e.name)); " *
+                            "a source must consume it before calling Tables.resolve"))
+    end
+    return e
+end
+
+"""
+    Tables.resolve(scan::Scan, names) -> BoundScan
+
+Resolve a `Scan` against a source's column names (any iterable of `Symbol`s).
```
nalimilan, line 463:
```suggestion
Resolve a [`Scan`](@ref) against a source's column names (any iterable of `Symbol`s).
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+# melts into per-element dynamic dispatch — measured 100× slower than the
+# same loop behind a barrier (62 ms → 0.6 ms over a 1M-row mask). The
+# barrier costs ONE dynamic dispatch per mask instead. The column kernels
+# inside `_evalexpr` specialize on the concrete container at their own call
+# boundary.
+@noinline _boolmask(v::AbstractVector) = Bool[x === true for x in v]
+"""
+    Tables.filtermask(scan_or_expr, table) -> AbstractVector{Bool}
+
+Evaluate a scan's filter over a table's columns: `mask[i]` is `true` iff row
```
nalimilan, line 583:
```suggestion
Evaluate a [Scan](@ref)'s filter over a table's columns:
`mask[i]` is `true` if and only if row
```

src/scan.jl
```diff
@@ -0,0 +1,723 @@
...
+_converted(::Nothing, c::AbstractVector) = c
+function _converted(::Type{T}, c::AbstractVector) where {T}
+    eltype(c) <: Union{T, Missing} && return c
+    anymissing = any(ismissing, c)
+    E = anymissing ? Union{T, Missing} : T
+    out = allocatecolumn(E, length(c))
+    @inbounds for i in eachindex(c)
+        x = c[i]
+        out[i] = ismissing(x) ? missing : convert(T, x)
+    end
```
nalimilan, line 611: Actually this should be enough given that `out`s eltype is `E`. This can be faster e.g. for `CategoricalArray`.
```suggestion
    copyto!(out, c)
```

src/scan.jl
quinnj: Good call; `DataAPI.All` is now reused in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; the operation codes are enums in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Yeah, I like that; comparisons now use `Tables.colcmp` in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; the docstring is now self-contained in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; `in_` is now `colin`, alongside `colcmp`, in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; this is now an internal identity check in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; the source-consumption contract is explicit in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; this now uses a fixed hardcoded seed in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; the error now names `ScanExpr` in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Yeah, I like `resolve`; changed in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; fixed in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; fixed in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; this is now `_findcols` in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; the wording is clearer in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; removed the unnecessary `Base` prefixes in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; the lifting semantics are explicit in the docs in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good catch; membership now preserves `missing` in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; `_boolmask` is now `@noinline` in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: Good call; this now uses `allocatecolumn` with a missing-safe loop in [64268c6](https://github.com/JuliaData/Tables.jl/commit/64268c6a316e380cc3da26965f440a5433ebc1f7).

src/scan.jl
quinnj: I initially tried making this `ismissing`, but it caused a crazy amount of invalidations, so switched it back to `isnull`.

src/scan.jl
quinnj: I prefer `validate`; other possible options that are simple would be like `raise`, or `raise_on_unmatched`.

quinnj:
> * Have you considered putting this in TableOperations.jl instead? Can this be used in interaction with functions from that package?

Thought about it, but this is kind of a different thing. I want to make it easy for Tables.jl sources to allow/facilitate pushdown support by taking a "Scan". That would require a lot of these packages adding a dep on TablesOperations, in addition to the Tables.jl dep they already have.


> * The name "scan" doesn't seem great to me. It's not super evocative outside of SQL, and in SQL ["full table scan"](https://en.wikipedia.org/wiki/Full_table_scan) refers to an operation which is generally slow as it goes over all rows without using an index. AFAICT a SQL-backed Tables.jl source could use an index under the hood e.g. if one wants to filter rows based on an index column, right? Also if you just use `Scan` to select and rename columns, no iteration over rows happens.

Yeah, it's not my favorite, but it's the terminology used in the broader Arrow project (Scanner), Iceberg (TableScan), and DataFusion (scan), which are some of the most up-to-date/modern libraries here. I don't really want to use `Query` because we already have all the queryverse stuff that could get confusing. There's `ReadPlan`, but that's kind of a mouthful and feels even more low-level SQL than scan IMO. `Projection` is a little linear-aglebraic. `Subset` would be my other serious consideration, but I worry it might be too specific for all the things it actually supports. Hence, scan feels like it has the best peer naming conventions in the industry.

nalimilan:
Ah I hadn't realized you wanted to tag a release so quickly. I have more comments, mostly about the docs. I'll finish them and post the review.

quinnj:
Ah, sorry for merging too soon. Trying to work out the last parts of Arrow.jl & CSV.jl rewrites and the Parsers.jl overhaul & this are upstream. Yeah, feel free to open an issue w/ thoughts or PR directly. Appreciate hte feedback/review!

src/scan.jl
nalimilan: What I mean is to have `Tables.ismissing` or `Tables.iscolmissing`, distinct from `Base.ismissing`.

src/scan.jl
```diff
@@ -0,0 +1,750 @@
...
+abstract type ScanExpr end
+
+struct Col <: ScanExpr
+    ref::Union{Symbol, String, Int}
+end
+"""
+    Tables.col(ref)
+
+A column reference inside a `Scan` filter: `col(:price) > 100`. Comparisons
+against literals, `colin`, `isnull`, `startswith`/`endswith`/`contains`, and
```
nalimilan, line 101:
```suggestion
against literals, [`colin`](@ref), [`isnull`](@ref), `startswith`/`endswith`/`contains`, and
```

test/scan.jl
```diff
@@ -0,0 +1,352 @@
...
+        @test out.a == [1, 2, 3, 4]
+        out = T.scan(nt, T.Scan(filter = T.colcmp(!=, T.col(:a), 1)))
+        @test out.a == [2, 3, 4]                                      # != never matches missing
+        out = T.scan(nt, T.Scan(filter = T.colin(T.col(:b), ("x", "w"))))
+        @test out.b == ["x", "x", "w"]
+        out = T.scan(nt, T.Scan(filter = startswith(T.col(:b), "z") | endswith(T.col(:b), "y")))
+        @test out.b == ["yy", "zzz"]
+        missingvals = (a = Union{Int, Missing}[1, missing, 3],
+                       b = Union{String, Missing}["x", missing, "z"])
+        @test T.filtermask(T.colin(T.col(:a), (1, missing)), missingvals) ==
```
nalimilan, line 150: Is there a test somewhere that `colin` returns `missing` when the column value is `missing`? AFAIK this test would pass even if it returned `false`. Same for `==`. Maybe it doesn't make a difference in practice but people may end up relying on it.

nalimilan (commented):
Here they are.

PR #381, opened 2026-08-26
Address post-merge Tables.Scan review feedback

quinnj:
## Summary

- route output conversion through `copyto!` so specialized column types can use optimized copying
- directly test that comparison and membership expressions preserve `missing` before mask lowering
- clarify the consumer and source documentation, including equality syntax, residual operations, and `OpNode` behavior
- centralize the `DataAPI.All` import and bump the patch version to 1.14.1

This follows up on Milan's late review of #380 while keeping the released Tables.Scan API unchanged.

## Validation

- `julia +1.9 --project=. -e 'using Pkg; Pkg.instantiate(); Pkg.test()'`
- `julia +1.12 --project=. -e 'using Pkg; Pkg.instantiate(); Pkg.test()'`
- `julia +1.12 --project=docs -e 'using Pkg; Pkg.instantiate(); include("docs/make.jl")'`

Co-authored by Codex

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tables"
 uuid = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.14.0"
+version = "1.14.1"
 
 [deps]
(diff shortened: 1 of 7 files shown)


nalimilan (approved):
Thanks!

PR #388, opened 2026-09-09
Preserve original values during column type inference

### LINKED ISSUE #387: `DataFrame(rows)`/`Tables.columntable` can silently corrupt values on table construction
`Tables.columntable` builds columns by incrementally widening their element type as new row types are seen, using `copyto!` from the previous column into a freshly allocated wider one. 

When a column needs two or more widenings before it stabilizes, each step copies from the *already-converted* intermediate column rather than the original rows, so precision lost in an early widening becomes permanent, even if the column ultimately widens to `Any`, which could have held the original value exactly. 

## Example 

```julia
julia> using Tables

julia> rows = [(; a = typemax(Int64)), (; a = 0.0), (; a = "")] # Create a tuple with an int, a float, and a string
3-element Vector{NamedTuple{(:a,)}}:
 (a = 9223372036854775807,)
 (a = 0.0,)
 (a = "",)

julia> ct = Tables.columntable(rows)
(a = Any[9.223372036854776e18, 0.0, ""],)

julia> ct.a[1] == rows[1].a
false

julia> using DataFrames

julia> DataFrame(rows).a
3-element Vector{Any}:
 9.223372036854776e18
 0.0
  ""

julia> DataFrame(rows).a[1] == rows[1].a
false
```

## Expected

```julia
julia> ct.a[1] == rows[1].a
true  # should preserve the original value exactly, not silently substitute a different Float64

julia> DataFrame(rows).a[1] == rows[1].a
true
```

quinnj:
Fixes #387.

Unknown-schema construction can round or reject a value during an intermediate conversion even when the final column could store it. For example, `typemax(Int64), 0.0, ""` must preserve the integer in the final `Any` column.

Ordinary column construction now keeps typed storage. When promotion would require conversion, storage widens to a union of the existing storage type and the incoming value type. The promoted output type is tracked separately, and conversion is deferred until that final type is known. Homogeneous columns keep their typed buffers. `@nospecialize` avoids excess value-type specialization in the insertion helper.

Dictionary construction keeps original values in column-aligned `Any` buffers, behind a function barrier. It unions column names and fills absent cells with `missing`. This also removes repeated full-column copying when integers follow floats. Both paths use the same final-column conversion helper.

The source is traversed once, including single-use inputs with a known length and inputs that reuse one mutable row. Final numeric promotion can still round values; this change prevents damage from intermediate conversions. Known-schema construction keeps its direct path.

### Performance

Apple M3 Max, Julia 1.12.7, BenchmarkTools 1.8.0, OrderedCollections 2.0.1, one Julia thread, 100,000 rows. Sequential processes with matching dependencies. Median milliseconds after warmup; up to 200 samples and a three-second budget. Main is `ee63bfb`, the old PR is `d464bfe`, and the update is `4fddf9f`.

| Case / constructor | Main | Old PR | Updated PR |
|---|---:|---:|---:|
| unknown_dense / `columntable` | 0.327 | 12.152 | 0.313 |
| unknown_stream / `columntable` | 0.583 | 12.533 | 0.615 |
| unknown_wide / `columntable` | 16.266 | 373.506 | 16.788 |
| late_widening / `columntable` | 3.947 | 8.080 | 2.672 |
| mixed_numeric / `columntable` | 4.218 | 18.357 | 2.346 |
| mixed_numeric / `dictcolumntable` | 4819.324 | 34.064 | 20.130 |

Main gives the wrong late-widening result. Its mixed-numeric dictionary result has one timed sample because one call exceeds the time budget; it allocates about 38.2 GiB through repeated copies, versus 3.85 MiB after this change. Dense homogeneous ordinary columns allocate 2.35 MiB, near main and down from 21.47 MiB in the old PR. Small timing differences are noise, not claimed improvements.

Fresh-process peak RSS at two million dense hidden-schema rows was 565.3 / 629.3 / 572.9 MiB for main / old PR / update. These are single-run process high-water marks. Fresh first-call tests at 1,000 rows added about 44–65 ms versus main on the two heterogeneous examples; costs vary with type combinations.

The benchmark script and reproduction instructions are in `benchmark/`. Docstrings and memory guidance describe typed widening separately from dictionary buffering.

### Validation

- Full `Pkg.test()` passes on Julia 1.9.4 and 1.12.7.
- All 172 focused checks pass, including the original 136.
- 1,800 deterministic mixed-type reference assertions pass.
- DataFrames 1.8.2 preserves all six issue #387 row permutations.
- Documentation and doctests build with Documenter 1.19.0.
- `git diff --check` passes.

Co-authored by Codex


docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
 of the input data; we just called [`Tables.rows`](@ref) or [`Tables.columns`](@ref) as was most natural for
 the table-specific use-case, knowing that it will Just Work™️.
 
+## Memory use when constructing columns
+
+The generic [`Tables.columns`](@ref) fallback and [`Tables.columntable`](@ref)
+construct in-memory columns from row sources. When the schema is unknown, they
+buffer the original cell values while inferring the final promoted column types,
+then convert each value directly to its final type. [`Tables.dictcolumntable`](@ref)
+uses the same buffer and also tracks columns that appear or disappear across rows.
```
nalimilan:
```suggestion
also uses a buffer and also tracks columns that appear or disappear across rows.
```
(not the same vector)

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+The source is traversed once, even when it reports a length: knowing the length
+does not guarantee that an iterator can be restarted safely.
+
+Peak memory includes the buffer, its bookkeeping, and the output columns. Buffering
+can also increase construction time. Known-schema row sources are written directly
```
nalimilan: Buffering is pretty cheap if the column entries turn out having a single type in the end, right? I wonder whether this presentation may make the fallback appear as slow, while it's pretty good in most typical cases.

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+column access use their own implementation.
+
+For larger workloads, choose an approach that matches the source and the work:
+
+- If the column types are known, provide a [`Tables.Schema`](@ref) from the row
+  source or construct typed column vectors directly. This avoids schema inference
+  and row buffering while still producing an in-memory table.
+- [DataFrames.jl](https://dataframes.juliadata.org/stable/lib/types/) is an in-memory
+  analysis package. `DataFrame(rows)` uses the Tables.jl column fallback, so changing
+  the sink alone does not avoid this buffer. With existing column vectors,
```
nalimilan: I don't get what "changing the sink" means. Using DataFrames instead of... what?

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+  `DataFrame(columns; copycols=false)` can avoid copying them; the DataFrame and
+  its source then share those vectors.
+- For data already stored in Arrow format,
+  [Arrow.jl](https://arrow.apache.org/julia/stable/manual/) supports memory-mapped
```
nalimilan, line 211: "supports" or rather "uses memory-mapped file access" (i.e. by default)?

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+  file access and `Arrow.Stream` iterates over record batches. Process batches
+  separately to avoid constructing one table containing every row. Writing an
+  arbitrary row source to Arrow first is not itself a guarantee of lower memory use.
```
nalimilan, line 214: When reading or when writing? Given mmapping, I would expect lower memory use when reading.

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+- For data that fits a SQL schema,
+  [SQLite.jl](https://juliadatabases.org/SQLite.jl/stable/) provides file-backed
+  databases, and [DuckDB](https://duckdb.org/docs/stable/connect/overview) supports
```
nalimilan, line 217:
```suggestion
  databases, and [DuckDB.jl](https://github.com/duckdb/DuckDB.jl) supports
```

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,48 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+  and row buffering while still producing an in-memory table.
+- [DataFrames.jl](https://dataframes.juliadata.org/stable/lib/types/) is an in-memory
+  analysis package. `DataFrame(rows)` uses the Tables.jl column fallback, so changing
+  the sink alone does not avoid this buffer. With existing column vectors,
+  `DataFrame(columns; copycols=false)` can avoid copying them; the DataFrame and
+  its source then share those vectors.
+- For data already stored in Arrow format,
+  [Arrow.jl](https://arrow.apache.org/julia/stable/manual/) supports memory-mapped
+  file access and `Arrow.Stream` iterates over record batches. Process batches
+  separately to avoid constructing one table containing every row. Writing an
```
nalimilan, line 213: Could also mention using the new scan feature to select rows, similar to what is said below about SQL?

src/dicts.jl
```diff
@@ -12,13 +12,20 @@ can be thought of as a `OrderedDict` mapping column names as `Symbol`s to `Abstr
...
+as opposed to taking the column names from the first row like `Tables.columns`. This
 means that as rows are iterated, each value from the row is joined into an aggregate
 final set of columns. This is especially useful when input table rows may not include
 columns if the value is missing, instead of including an actual value `missing`, which
 is common in json, for example. This results in a performance cost tracking all seen
 values and inferring the final unioned schemas, so it's recommended to use only when
 needed.
+
+For unknown schemas, the source is read once and the original cell values are buffered
+while column types are inferred with `promote_type`. Final columns are then allocated
```
nalimilan:
```suggestion
while column types are widened if necessary with `promote_type`. Final columns are then allocated
```

src/dicts.jl
```diff
@@ -12,13 +12,20 @@ can be thought of as a `OrderedDict` mapping column names as `Symbol`s to `Abstr
...
+and each value is converted directly to its final column type. Absent cells become
+`missing`. Peak memory includes both the buffered values and output columns, plus
+buffering overhead. Mutable cell contents are not copied. Final numeric promotion can
+still round values. Rows with a known schema do not need this intermediate buffer.
```
nalimilan:
```suggestion
still round values. Tables with a known schema do not need this intermediate buffer.
```

src/dicts.jl
```diff
@@ -12,13 +12,20 @@ can be thought of as a `OrderedDict` mapping column names as `Symbol`s to `Abstr
...
 final set of columns. This is especially useful when input table rows may not include
 columns if the value is missing, instead of including an actual value `missing`, which
 is common in json, for example. This results in a performance cost tracking all seen
 values and inferring the final unioned schemas, so it's recommended to use only when
 needed.
+
+For unknown schemas, the source is read once and the original cell values are buffered
+while column types are inferred with `promote_type`. Final columns are then allocated
+and each value is converted directly to its final column type. Absent cells become
+`missing`. Peak memory includes both the buffered values and output columns, plus
```
nalimilan:
```suggestion
`missing`. For columns whose elements are all of the same type,
the buffer becomes the final column vector, without copying.
Peak memory includes both the buffered values and output columns, plus
```

src/fallbacks.jl
```diff
@@ -149,53 +149,62 @@ end
...
 
-function _buildcolumns(rowitr, row, st, sch, columns, updated)
-    eachcolumns(add_or_widen!, sch, row, columns, 1, updated, Base.IteratorSize(rowitr))
-    return __buildcolumns(rowitr, st, sch, updated[], 1, updated)
+function padmissing!(values, len)
+    start = length(values) + 1
+    resize!(values, len)
+    for i in start:len
+        values[i] = missing
+    end
```
nalimilan: Shorter, and could be slightly faster when an optimized method exists (e.g. CategoricalArrays):
```suggestion
    values[start:len] .= missing
```

nalimilan:
Hmm, the huge performance regression for unknown types with `columntable` is problematic IMO. At least it should be possible to avoid it for homogeneous columns by continuing to allocate a buffer with an eltype equal to the type of the first element, and only if `promote_type` returns a type which isn't a supertype of that, allocate a new buffer with eltype `Any` (or even using `typejoin`, as we want to be fast for `Union{T, Missing}`)? It's trickier but the machinery is already in place and I think it's worth it.

Your suggestion at https://github.com/JuliaData/Tables.jl/issues/387#issuecomment-5593879487 to make two passes over data with known length is also interesting, as it would probably be much faster and avoid allocating large buffers. It could be applied only to `AbstractVector` inputs if the API doesn't allow doing that in general.

quinnj:
Yeah, I agree that the original `columntable` regression was too large. I've pushed a rework based on your first suggestion.

Ordinary columns now start with typed storage. If promotion would require converting existing values, storage widens to a union instead. The promoted output type is tracked separately, and conversion happens only once that final type is known. This keeps homogeneous columns fast and preserves small unions such as `Union{Int, Missing}` and `Union{Int, Float64}`. It also handles the reverse order: an int arriving after floats must not be rounded early either.

Fresh one-thread benchmarks on Julia 1.12, at 100,000 rows, give these median times for `columntable`:

| Case | Main | Previous PR | Updated PR |
|---|---:|---:|---:|
| Hidden-schema dense | 0.327 ms | 12.152 ms | 0.313 ms |
| 32 columns | 16.266 ms | 373.506 ms | 16.788 ms |
| Alternating Int/Float64 | 4.218 ms | 18.357 ms | 2.346 ms |
| Late widening from #387 | 3.947 ms* | 8.080 ms | 2.672 ms |

*Main produces the wrong values in the late-widening case. Dense-column allocation is back near main. One remaining trade-off is compilation: the fresh heterogeneous examples took about 44–65 ms more on their first call than main.

I also tested the vector-only two-pass prototype and switching to `Any` on mismatch. Both were slower than union storage on the mixed-numeric case. That does not rule out a better two-pass implementation, but I kept one general source traversal for this change.

For `dictcolumntable`, I kept the `Any` buffers behind a function barrier. This also removes the existing repeated-copy problem when ints follow floats.

quinnj:
[Reviewed by AI]

Reviewed 4fddf9f, focusing on lossless intermediate storage, final promotion, sparse dictionary columns, and single-use/reused-row inputs. The full Julia 1.12.6 suite passes, including all 172 buffering checks. I found no blocking issue in that scope. The documentation correctly distinguishes intermediate-value preservation from possible rounding during final numeric promotion. I did not independently repeat the published performance/RSS measurements or validate other Julia versions.

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,54 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+the final promoted type is known. Homogeneous columns keep their typed storage.
+[`Tables.dictcolumntable`](@ref) instead buffers cells in `Any` vectors and tracks
+columns that appear or disappear across rows. Both paths convert values directly to
+their final column types, so intermediate numeric promotion cannot lose information.
+Final numeric promotion can still round values, such as a large integer converted to
+`Float64`. Mutable cell contents are not copied.
+The source is traversed once, even when it reports a length: knowing the length
+does not guarantee that an iterator can be restarted safely.
+
+For ordinary columns, widening and final conversion can temporarily require both old
```
nalimilan, line 194: What does "ordinary columns" mean exactly?

Also:
```suggestion
For ordinary columns, when the schema is unknown, widening and final conversion can temporarily require both old
```

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,54 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+and new storage. Dictionary construction can retain `Any` buffers alongside the output
+columns, plus bookkeeping. Known-schema row sources are written directly to typed
+columns without inference or intermediate buffers. Sources that already provide column
+access use their own implementation.
+
+For larger workloads, choose an approach that matches the source and the work:
+
+- If the column types are known, provide a [`Tables.Schema`](@ref) from the row
+  source or construct typed column vectors directly. This avoids schema inference
+  and intermediate storage while still producing an in-memory table.
```
nalimilan, line 204:
```suggestion
  and intermediate storage if the column has a heterogeneous eltype,
  while still producing an in-memory table.
```

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,54 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+[`Tables.dictcolumntable`](@ref) instead buffers cells in `Any` vectors and tracks
+columns that appear or disappear across rows. Both paths convert values directly to
+their final column types, so intermediate numeric promotion cannot lose information.
+Final numeric promotion can still round values, such as a large integer converted to
+`Float64`. Mutable cell contents are not copied.
+The source is traversed once, even when it reports a length: knowing the length
+does not guarantee that an iterator can be restarted safely.
+
+For ordinary columns, widening and final conversion can temporarily require both old
+and new storage. Dictionary construction can retain `Any` buffers alongside the output
```
nalimilan, line 195:
```suggestion
and new storage. Dictionary construction can temporarily use `Any` buffers alongside the output
```

src/dicts.jl
```diff
@@ -12,13 +12,20 @@ can be thought of as a `OrderedDict` mapping column names as `Symbol`s to `Abstr
...
+as opposed to taking the column names from the first row like `Tables.columns`. This
 means that as rows are iterated, each value from the row is joined into an aggregate
 final set of columns. This is especially useful when input table rows may not include
 columns if the value is missing, instead of including an actual value `missing`, which
 is common in json, for example. This results in a performance cost tracking all seen
 values and inferring the final unioned schemas, so it's recommended to use only when
 needed.
+
+For unknown schemas, the source is read once and the original cell values are buffered
+in `Any` vectors while column types are inferred with `promote_type`. Values are then
```
nalimilan, line 24:
```suggestion
in `Any` vectors while column types are chosen with `promote_type`. Values are then
```

src/dicts.jl
```diff
@@ -12,13 +12,20 @@ can be thought of as a `OrderedDict` mapping column names as `Symbol`s to `Abstr
...
+converted directly to their final column types. Absent cells become `missing`. Peak
+memory can include both the buffered values and output columns, plus buffering overhead.
+Mutable cell contents are not copied. Final numeric promotion can still round values.
+Rows with a known schema do not need this intermediate buffer.
```
nalimilan, line 28:
```suggestion
Tables with a known schema do not need this intermediate buffer.
```

src/dicts.jl
```diff
@@ -39,54 +46,55 @@ function dictcolumntable(x)
...
             sch = Schema(collect(keys(out)), eltype.(values(out)))
         end
     end
     return DictColumnTable(sch, out)
 end
 
+# Column names and types are only known once every row is seen, so buffer the original
+# values and convert each one once into its final column type. Kept in its own function
+# so the hot loop's locals stay type-stable.
+function buffercolumns(r, len)
```
nalimilan, line 60:
```suggestion
function buffercolumns(r, len::Integer)
```

src/dicts.jl
```diff
@@ -39,54 +46,55 @@ function dictcolumntable(x)
...
+            end
+            vals = buffers[i]
+            T = types[i]
+            if length(vals) < n - 1 # absent from earlier rows
+                append!(vals, Iterators.repeated(missing, n - 1 - length(vals)))
+                T = Union{Missing, T}
+            end
+            val = getcolumn(row, nm)
+            push!(vals, val)
+            types[i] = val isa T ? T : promote_type(T, typeof(val))
```
nalimilan, line 84: Probably a bit faster?
```suggestion
            val isa T || (types[i] = promote_type(T, typeof(val)))
```

src/fallbacks.jl
```diff
@@ -152,33 +152,42 @@ end
...
+    V = typeof(val)
+    P2 = promote_type(P, V)
+    S = T <: P2 && V <: P2 ? P2 : Union{T, V}
+    new = allocatecolumn(S, length(dest))
+    nfilled > 0 && copyto!(new, 1, dest, 1, nfilled)
+    return new, P2
+end
+
+# Convert a column once into its promoted type if storage was widened losslessly instead.
+finishcolumn(col::AbstractVector{T}, ::Type{P}) where {T, P} = T === P ? col : copyto!(allocatecolumn(P, length(col)), col)
```
nalimilan, line 168:
```suggestion
finishcolumn(col::AbstractVector{T}, ::Type{P}) where {T, P} =
    T === P ? col : copyto!(allocatecolumn(P, length(col)), col)
```

src/fallbacks.jl
```diff
@@ -152,33 +152,42 @@ end
 @inline add!(dest::AbstractArray, val, ::Union{Base.HasLength, Base.HasShape}, row) = setindex!(dest, val, row)
 @inline add!(dest::AbstractArray, val, T, row) = push!(dest, val)
 
-@inline function add_or_widen!(val, col::Int, nm, dest::AbstractArray{T}, row, updated, L) where {T}
-    if val isa T || promote_type(typeof(val), T) <: T
+# Copy the filled values into storage that also holds `val` without conversion.
+# `P` tracks the promoted output type. If promotion is not a supertype of both storage
+# and the new value, use their union and defer conversion until the final type is known.
+function widen(dest::AbstractArray{T}, ::Type{P}, val, nfilled) where {T, P}
```
nalimilan, line 158: Call this `widencolumn` to avoid confusion with `Base.widen` and for consistency with related functions?

src/namedtuples.jl
```diff
@@ -150,6 +150,15 @@ also known as a "column table". A "column table" is a kind of default
 table type of sorts, since it satisfies the Tables.jl column interface
 naturally.
 
+When constructing columns from rows with an unknown schema, this function reads the
+source once and starts each column with the first value's type. Storage widens without
+converting existing values, using a union when numeric promotion could lose information.
+Column types are inferred with `promote_type`; any required conversion is deferred until
```
nalimilan, line 156:
```suggestion
Column types are chosen with `promote_type`; any required conversion is deferred until
```

src/namedtuples.jl
```diff
@@ -150,6 +150,15 @@ also known as a "column table". A "column table" is a kind of default
...
+the final types are known. Widening and final conversion can temporarily require both
+old and new column storage. Mutable cell contents are not copied. Final numeric promotion
+can still round values, such as a large integer converted to `Float64`. Rows with a known
+schema are written directly to typed columns without this inference step.
```
nalimilan, line 160:
```suggestion
schema or with a homogeneous element type are written directly to typed columns
without intermediate buffers.
```

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,54 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+columns, plus bookkeeping. Known-schema row sources are written directly to typed
+columns without inference or intermediate buffers. Sources that already provide column
```
nalimilan, line 197: Inference isn't really involved here as the actual type of values is used.
```suggestion
columns without intermediate buffers. Sources that already provide column
```

docs/src/using-the-interface.md
```diff
@@ -176,6 +176,54 @@ Note in both the rows and columns usages, we didn't need to worry about the natu
...
+access use their own implementation.
+
+For larger workloads, choose an approach that matches the source and the work:
+
+- If the column types are known, provide a [`Tables.Schema`](@ref) from the row
+  source or construct typed column vectors directly. This avoids schema inference
```
nalimilan, line 203:
```suggestion
  source or construct typed column vectors directly. This avoids schema detection
```

test/buffering.jl
```diff
@@ -0,0 +1,104 @@
...
+        for p in ((1,2,3), (1,3,2), (2,1,3), (2,3,1), (3,1,2), (3,2,1))
+            input = rows[collect(p)]
+            for source in (input, Iterators.filter(_ -> true, input),
+                           BufferedTestRows(input), BufferedTestRows(input, Base.HasLength()))
+                col = constructor(source).a
+                @test eltype(col) === Any
+                @test all(col[i] === input[i].a for i in eachindex(input))
+            end
+        end
+        # Intermediate conversion can throw, even when the final type is Any.
```
nalimilan, line 42:
```suggestion
        # Intermediate conversion can throw, even when the final type is Any (#387)
```

nalimilan (commented):
Nice!

PR #389, opened 2026-09-19
Honor type-level materializers for custom row tables

### LINKED ISSUE #354: materializer behavior differs from its docs

quinnj:
A custom `AbstractVector` of named tuples can define the documented `Tables.materializer(::Type{<:MyTable})` method, but `Tables.materializer(instance)` still returns `rowtable`. The instance-level `RowTable` fallback bypasses the generic delegation to the type.

Move that fallback to the type level. Ordinary row vectors and views still use `rowtable`, while custom row tables can supply their own sink through the documented extension point. Add seven checks for ordinary, empty, viewed, and custom row tables.

Fixes #354.

Validation: the custom-row-table reproduction fails on main. Full test suites pass on Julia 1.9.4 and 1.12.6.

Co-authored by Codex


quinnj (commented):
[Reviewed by AI]

Reviewed 235ef8c. No actionable findings in the materializer change. An independent custom row-vector reproducer fails on main and passes here; standard row/column sinks remain correct. The full Julia 1.12 suite passed 466,379 checks, plus 12 focused checks. No materializer dispatch ambiguities were found.

nalimilan (approved):

PR #390, opened 2026-09-19
Apply wide-schema fallback to tuple-based constructors

### LINKED ISSUE #331: Inconsistencies in `Schema`
Generically, wide table schemas will bypass the type system:

https://github.com/JuliaData/Tables.jl/blob/2cb13998e856692ed273c931b83477caf8b7b020/src/Tables.jl#L477-L483

But not for `MatrixTable`s:

https://github.com/JuliaData/Tables.jl/blob/2cb13998e856692ed273c931b83477caf8b7b020/src/matrix.jl#L40

Maybe the branch should be put in the constructor, such that it is done for any table type?

C.f. https://github.com/apache/arrow-julia/pull/415

quinnj:
A matrix with 65,536 columns returns a schema whose names and types remain encoded in its type, although the general schema constructor switches to stored vectors beyond 65,535 columns. The tuple/type constructor bypasses that existing limit.

Apply the existing threshold in the shared tuple/type constructor and route the NamedTuple-type constructor through it. This covers matrix rows and columns as well as direct schema construction. Narrow schemas keep their current specialized representation. Add 14 checks for the wide and narrow cases.

Fixes #331.

Validation: the wide-matrix reproduction fails on main (`stored=false`). Full test suites pass on Julia 1.9.4 and 1.12.6.

Co-authored by Codex


quinnj (commented):
[Reviewed by AI] No actionable findings at c175696. Checked both sides of the 65,535-column boundary, both constructors, and `eachcolumn` on an actual 65,536-column matrix. All 35 independent checks and the full Julia 1.12.6 suite (466,386 checks) passed. Narrow schemas retain their specialized representation. Local validation was on macOS arm64.

src/Tables.jl
```diff
@@ -470,8 +470,12 @@ struct Schema{names, types}
 end
 
 Schema{names, types}() where {names, types} = Schema{names, types}(nothing, nothing)
-Schema(names::Tuple{Vararg{Symbol}}, ::Type{T}) where {T <: Tuple} = Schema{names, T}()
-Schema(::Type{NamedTuple{names, types}}) where {names, types} = Schema{names, types}()
+function Schema(names::Tuple{Vararg{Symbol}}, ::Type{T}) where {T <: Tuple}
+    length(names) > SCHEMA_SPECIALIZATION_THRESHOLD &&
+        return Schema(names, (fieldtype(T, i) for i in 1:fieldcount(T)))
+    return Schema{names, T}()
```
```suggestion
    if length(names) > SCHEMA_SPECIALIZATION_THRESHOLD
        return Schema(names, (fieldtype(T, i) for i in 1:fieldcount(T)))
    else
        return Schema{names, T}()
    end
```

nalimilan (approved):

